PR #23, opened 2024-12-20
feat: add tenant and env module with test cases

awsmankit:


nkmishra1997:
Please improve the title of the PR. Suggestions : `Add tenant module and improve test cases`

internal/domain/tenant/repository.go
```diff
@@ -0,0 +1,10 @@
+package tenant
+
+import (
+	"context"
+)
+
+type Repository interface {
+	Create(ctx context.Context, tenant *Tenant) error
+	GetById(ctx context.Context, id string) (*Tenant, error)
+}
```
nkmishra1997: We need to implement these repositories as well in `/internal/repository/postgres/tenant.go`

internal/domain/tenant/repository.go
```diff
@@ -0,0 +1,10 @@
+package tenant
+
+import (
+	"context"
+)
+
+type Repository interface {
+	Create(ctx context.Context, tenant *Tenant) error
+	GetById(ctx context.Context, id string) (*Tenant, error)
```
nkmishra1997:
```suggestion
	GetByID(ctx context.Context, id string) (*Tenant, error)
```

We are using `ID` for `id` fields everywhere so lets make it consistent

internal/api/v1/tenant.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func NewTenantHandler(tenantService *service.TenantService) *TenantHandler {
+	return &TenantHandler{tenantService: tenantService}
+}
+
+func (h *TenantHandler) CreateTenant(c *gin.Context) {
+	var req dto.CreateTenantRequest
+	if err := c.ShouldBindJSON(&req); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
```
nkmishra1997: use `ErrorResponse` as defined in `internal/api/v1/errors.go`

internal/api/v1/tenant.go
```diff
@@ -0,0 +1,47 @@
...
+
+type TenantHandler struct {
+	tenantService *service.TenantService
+}
+
+func NewTenantHandler(tenantService *service.TenantService) *TenantHandler {
+	return &TenantHandler{tenantService: tenantService}
+}
+
+func (h *TenantHandler) CreateTenant(c *gin.Context) {
```
nkmishra1997, line 32: Add annotations so that it is available in the API specs refer any api route and see the pattern in comments example `api/v1/wallet.go`

internal/service/plan_test.go
```diff
@@ -0,0 +1,166 @@
...
+	planService *planService
+	planRepo    *testutil.InMemoryPlanRepository
+}
+
+func TestPlanService(t *testing.T) {
+	suite.Run(t, new(PlanServiceSuite))
+}
+
+func (s *PlanServiceSuite) SetupTest() {
+	s.ctx = context.Background()
```
nkmishra1997: use `testutils.SetupContext()`

internal/service/price_test.go
```diff
@@ -0,0 +1,126 @@
...
+	// Non-existent price
+	resp, err = s.priceService.GetPrice(s.ctx, "nonexistent-id")
+	s.Error(err)
+	s.Nil(resp)
+}
+
+func (s *PriceServiceSuite) TestGetPrices() {
+	// Prepopulate the repository with prices
+	_ = s.priceRepo.Create(s.ctx, &price.Price{ID: "price-1", Amount: decimal.NewFromInt(100), Currency: "USD"})
+	_ = s.priceRepo.Create(s.ctx, &price.Price{ID: "price-2", Amount: decimal.NewFromInt(200), Currency: "USD"})
```
nkmishra1997: ideally we should not be able to create prices without plan_id so lets validate that as well and even dummy data should reflect that

awsmankit: this i have not tested yet will fixup in next commit

internal/domain/tenant/repository.go
awsmankit: done with this

migrations/postgres/V9__create_environment_table.sql
```diff
@@ -0,0 +1,14 @@
+---- Create environmnet table ----
+
+CREATE TABLE IF NOT EXISTS environments (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    name VARCHAR(255) NOT NULL UNIQUE,
```
nkmishra1997: no need to be unique

migrations/postgres/V9__create_environment_table.sql
```diff
@@ -0,0 +1,14 @@
...
+    type VARCHAR(50) NOT NULL,
+    slug VARCHAR(255) NOT NULL UNIQUE,
```
nkmishra1997: no need to be unique. IT must be unique across a tenant not across all tenants so we can put a constraint on tenant_id+slug maybe

internal/api/v1/environment.go
```diff
@@ -0,0 +1,129 @@
...
+// @Summary Create an environment
+// @Description Create an environment
+// @Tags environments
+// @Accept json
+// @Produce json
+// @Security BearerAuth
+// @Param environment body dto.CreateEnvironmentRequest true "Environment"
+// @Success 201 {object} dto.EnvironmentResponse
+// @Failure 400 {object} errors.ErrorResponse
+// @Failure 500 {object} errors.ErrorResponse
```
nkmishra1997: I think `ErrorResponse`  is already in `api/v1` and not a separate package so we need to use just `ErrorResponse` here

migrations/postgres/V9__create_environment_table.sql
nkmishra1997: For now we can just remove the unique check here and keep it in the service layer

internal/api/v1/environment.go
nkmishra1997: Once fixed - do run `make swagger` to generate the API docs as well for these APIs

internal/api/v1/tenant.go
```diff
@@ -0,0 +1,69 @@
...
+// @Summary Create a new tenant
+// @Description Create a new tenant
+// @Tags Tenant
+// @Accept json
+// @Produce json
+// @Security BearerAuth
+// @Param request body dto.CreateTenantRequest true "Create tenant request"
+// @Success 201 {object} dto.TenantResponse
+// @Failure 400 {object} environmentErrorResponse
+// @Failure 500 {object} environmentErrorResponse
```
nkmishra1997: where is this environmentErrorResponse defined?

internal/domain/environment/model.go
```diff
@@ -0,0 +1,20 @@
...
+	types.BaseModel
+}
+
+type EnvironmentType string
+
+const (
+	EnvironmentDevelopment EnvironmentType = "development"
+	EnvironmentTesting     EnvironmentType = "testing"
+	EnvironmentProduction  EnvironmentType = "production"
+)
```
nkmishra1997: move these to `internal/types` as we are defining all types there so that types can be used everywhere in the code without having to import domains specifically. This helps in avoiding import cycles in future (hypothetical ex say subscription model needs `invoice_type` and invoice model needs `subscription_type` then in that case it will lead to import cycles.

internal/repository/postgres/environment.go
```diff
@@ -0,0 +1,86 @@
...
+
+	r.logger.Debug("creating environment", "environment_id", env.ID)
+
+	_, err := r.db.NamedExecContext(ctx, query, env)
+	return err
+}
+
+func (r *environmentRepository) Get(ctx context.Context, id string) (*environment.Environment, error) {
+	var env environment.Environment
+	query := `SELECT * FROM environments WHERE id = :id`
```
nkmishra1997: add tenant_id filter here (need to add in all queries)

internal/repository/postgres/environment.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get environment: %w", err)
+	}
+	return &env, nil
+}
+
+func (r *environmentRepository) List(ctx context.Context, filter types.Filter) ([]*environment.Environment, error) {
+	var environments []*environment.Environment
+	query := `
+		SELECT * FROM environments ORDER BY created_at DESC LIMIT :limit OFFSET :offset`
```
nkmishra1997: same here add tenant_id filter

internal/repository/postgres/environment.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		environments = append(environments, &env)
+	}
+
+	return environments, nil
+}
+
+func (r *environmentRepository) Update(ctx context.Context, env *environment.Environment) error {
+	query := `
+		UPDATE environments SET
```
nkmishra1997, line 87: here as well we need tenant_id check

migrations/postgres/V9__create_environment_table.sql
awsmankit: fixed

migrations/postgres/V9__create_environment_table.sql
awsmankit: fixed

internal/api/v1/tenant.go
awsmankit: it should be Error Response : fixed

internal/domain/environment/model.go
awsmankit: done

nkmishra1997 (approved):

PR #33, opened 2024-12-27
added update meter api 

awsmankit:


internal/config/config.yaml
```diff
@@ -5,7 +5,7 @@ server:
   address: ":8080"
 
 auth:
-  provider: "flexprice" # "flexprice" or "supabase"
+  provider: "supabase" # "flexprice" or "supabase"
```
nkmishra1997: don't change this

internal/repository/postgres/meter.go
```diff
@@ -193,3 +193,83 @@ func (r *meterRepository) DisableMeter(ctx context.Context, id string) error {
...
+	if len(existingFiltersJSON) > 0 {
+		if err := json.Unmarshal(existingFiltersJSON, &existingFilters); err != nil {
+			return fmt.Errorf("unmarshal existing filters: %w", err)
+		}
+	}
+
+	// Merge new filters into existing filters
+	filterMap := map[string][]string{}
+	for _, f := range existingFilters {
+		filterMap[f.Key] = f.Values
```
nkmishra1997, line 229: All of the business logic of what to update and what not to update should be in service layer and not in repository layer.
Here we should have just one responsibility - to execute DB query with right set of validations and control.

internal/repository/postgres/meter.go
```diff
@@ -193,3 +193,83 @@ func (r *meterRepository) DisableMeter(ctx context.Context, id string) error {
...
+			Key:    key,
+			Values: values,
+		})
+	}
+
+	updatedFiltersJSON, err := json.Marshal(updatedFilters)
+	if err != nil {
+		return fmt.Errorf("marshal updated filters: %w", err)
+	}
+
```
nkmishra1997: All of the business logic of what to update and what not to update should be in service layer and not in repository layer.
Here we should have just one responsibility - to execute DB query with right set of validations and control.

internal/service/auth.go
```diff
@@ -52,14 +52,14 @@ func (s *authService) SignUp(ctx context.Context, req *dto.SignUpRequest) (*dto.
 		return nil, fmt.Errorf("unable to sign up: %w", err)
 	}
 
-	auth := auth.NewAuth(user.ID, types.AuthProviderFlexprice, authResponse.ProviderToken)
```
nkmishra1997: these changes are not required

internal/service/auth.go
```diff
@@ -52,14 +52,14 @@ func (s *authService) SignUp(ctx context.Context, req *dto.SignUpRequest) (*dto.
...
+	auth := auth.NewAuth(user.ID, types.AuthProviderSupabase, authResponse.ProviderToken)
 	err = s.authRepo.CreateAuth(ctx, auth)
 	if err != nil {
 		return nil, fmt.Errorf("unable to create auth: %w", err)
 	}
 
 	response := &dto.AuthResponse{
-		Token: authResponse.AuthToken,
+		Token: authResponse.ProviderToken,
```
nkmishra1997: all changes in this file are not required

internal/repository/postgres/meter.go
nkmishra1997: And rather than doing raw queries - when you implement this logic in service layer - use service methods like GetMeter which handle all the json parsing etc out of the box

go.mod
```diff
@@ -32,6 +32,11 @@ require (
 	golang.org/x/time v0.8.0
 )
 
+require (
```
nkmishra1997, line 39: pls run `go mod tidy` once

internal/service/meter.go
```diff
@@ -63,3 +64,27 @@ func (s *meterService) DisableMeter(ctx context.Context, id string) error {
...
+		return nil, fmt.Errorf("update meter: %w", err)
+	}
+
+	// Fetch the updated meter
+	updatedMeter, err := s.meterRepo.GetMeter(ctx, id)
+	if err != nil {
+		return nil, fmt.Errorf("get meter: %w", err)
+	}
+
+	return updatedMeter, nil
```
nkmishra1997: don't fetch right after writing updating the item in DB
return the edited object

internal/service/meter.go
```diff
@@ -63,3 +64,75 @@ func (s *meterService) DisableMeter(ctx context.Context, id string) error {
 	}
 	return s.meterRepo.DisableMeter(ctx, id)
 }
+
+// contains checks if a slice contains a specific value
+func contains(slice []string, value string) bool {
```
nkmishra1997, line 69: Lets use https://github.com/samber/lo?tab=readme-ov-file#contains

internal/service/meter.go
```diff
@@ -63,3 +64,75 @@ func (s *meterService) DisableMeter(ctx context.Context, id string) error {
...
+	mergedFilters := mergeFilters(existingMeter.Filters, filters)
+
+	// Update only the filters field in the database
+	if err := s.meterRepo.UpdateMeter(ctx, id, mergedFilters); err != nil {
+		return nil, fmt.Errorf("update filters: %w", err)
+	}
+
+	// Return the updated meter object
+	existingMeter.Filters = mergedFilters
+	return existingMeter, nil
```
nkmishra1997, line 104: first update the existingMeter object
`existingMeter.Filters = mergedFilters`
then pass the entire meter object to `s.meterRepo.UpdateMeter(ctx, existingMeter)`

nkmishra1997 (approved):

PR #34, opened 2024-12-28
feat(invoice): set up invoicing

nkmishra1997:


internal/api/dto/invoice.go
```diff
@@ -0,0 +1,174 @@
...
+	Metadata       map[string]interface{}      `json:"metadata,omitempty"`
+}
+
+func (r *CreateInvoiceRequest) Validate() error {
+	validate := validator.New()
+	if err := validate.Struct(r); err != nil {
+		return err
+	}
+
+	if r.AmountDue.IsZero() || r.AmountDue.IsNegative() {
```
nkmishra1997, line 37: Should we allow $0 invoices?

PR #53, opened 2025-01-14
added system event service and temporal workflows

HemabhKamboj:
# Add Temporal Worker for Billing Period Updates

Adds Temporal workflow support for handling billing period updates. This introduces a new worker mode and sets up the necessary workflow infrastructure for reliable billing period processing.
### Temporal Integration
- Added Temporal client and worker setup
- Implemented `UpdateBillingPeriodsWorkflow` and activity
- Added `temporal_worker` deployment mode

### System Event Service
- Added new event type `SystemEventTypeUpdateBillingPeriods`
- Updated system event service to trigger Temporal workflows
- Added workflow options with daily cron schedule

## Notes
- Testing is in progress
- Workflow runs daily at midnight via cron schedule
- Uses existing DI patterns with fx
- System events are tracked and status is updated after workflow execution


ent/schema/systemevent.go
```diff
@@ -0,0 +1,49 @@
...
+type SystemEvent struct {
+	ent.Schema
+}
+
+// Fields of the SystemEvent.
+func (SystemEvent) Fields() []ent.Field {
+	return []ent.Field{
+		field.UUID("id", uuid.UUID{}).
+			Default(uuid.New).
+			Unique(),
```
nkmishra1997, line 20: We have slightly changed how we define schema, primary keys, etc pls refer https://github.com/flexprice/flexprice/blob/develop/ent/schema/invoice.go#L20

PR #88, opened 2025-02-16
wip error handler with cockroachdb

BRO3886:
Used cockroachDB for error handling. Unwrapped to get root cause. Coupled status code with error.

Implemented a middleware called error handler. This uses gin's internal error messages. Errors are registered to gin context via `c.Error()` and all error processing is done via the middleware. We could make it more verbose in the future, and also standardise error message format this way.

Currently I used safe details as a carrier to get the err code. Will explore what all can be done with this

> Coupled status code with error

Initially I thought it's a nice idea but rn getting 2nd thoughts about this


internal/domain/tenant/errors.go
```diff
@@ -0,0 +1,11 @@
+package tenant
+
+import (
+	"github.com/cockroachdb/errors"
+	ierr "github.com/flexprice/flexprice/internal/errors"
+)
+
+var (
+	ErrNotFound      = errors.WithSafeDetails(errors.New("tenant not found"), ierr.ErrCodeNotFound)
+	ErrAlreadyExists = errors.WithSafeDetails(errors.New("tenant already exists"), ierr.ErrCodeAlreadyExists)
```
nkmishra1997: representational

internal/rest/middleware/errhandler.go
```diff
@@ -0,0 +1,30 @@
...
+		if len(c.Errors) > 0 {
+			err := c.Errors.Last().Err
+
+			// Create error response
+			errResp := gin.H{
+				"success": false,
+				"error": gin.H{
+					"message": errors.UnwrapAll(err).Error(),
+				},
+			}
```
nkmishra1997, line 23: I think we will have to suport something like a display_message field so that frontend can also show business validatio errors if required.
If not then maybe codes so that Frontend can handle the error display logic at their end. How will we accomplish this?

The middleware is solving for error meessgae, chainging and proper HTTP status - so that part looks great.

internal/repository/postgres/tenant.go
```diff
@@ -17,26 +19,43 @@ func NewTenantRepository(db *postgres.DB, logger *logger.Logger) tenant.Reposito
...
+		query,
+		t.Name,
+	).Scan(&exists)
+	if err != nil && err != sql.ErrNoRows {
+		return err
+	}
+
+	if exists {
+		return ierr.New(ierr.ErrAlreadyExists, "tenant already exists")
+	}
```
nkmishra1997, line 38: Although the change is fine - would recommend to not pick these improvements on the way
We already have a PR open for moving tenant, user and env domains to ent so it handles all these things out fo the box

internal/httpclient/error.go
```diff
@@ -2,31 +2,24 @@ package httpclient
...
-	return e.InternalError.Unwrap()
-}
-
 func (e *Error) Error() string {
-	return e.InternalError.Error()
+	return fmt.Sprintf("%d - %v", e.StatusCode, e.Response)
 }
 
 // NewError creates a new HTTP client error
 func NewError(statusCode int, response []byte) *Error {
```
nkmishra1997, line 19: Do we still need a specific error for the http module layer?

internal/rest/middleware/errhandler.go
BRO3886: > I think we will have to suport something like a display_message field so that frontend can also show business validatio errors if required

thinking about an approach for this

internal/repository/postgres/tenant.go
BRO3886: got it

internal/httpclient/error.go
BRO3886: kept it for `func shouldRetry(logger *logger.Logger, err error) bool {` - this takes action based on the response status, but this function is not being called as of now

PR #97, opened 2025-02-21
Error builder with cockroachdb/errors

BRO3886:
This PR introduces the new error api based on the discussion https://github.com/flexprice/flexprice/discussions/80.

## Intro

I propose deprecating the internal errors struct in favour of the extensive cockroachDB errors package. It's a drop in replacement for golang's own error package. Here I have created an error builder for the sake of developer convenience, which internally makes use of methods provided by [cockroachDB/errors](https://github.com/cockroachdb/errors) - such as `Mark()`, `WithSafeDetails()` etc.

This PR also introduces a middleware - `ErrHandler` which can be used for all kinds of error handling on the REST API. Instead of `NewErrorResponse` we can just use `c.Error(err)` and the middleware will handle the error messaging (response).

## Features / How to use the `ErrorBuilder`

> Mark() should be the last method that is called on the error builder. This is intentional to ensure all errors can be mapped to a status code.

Wrapping existing errors:
```go
if err := r.client.Querier(ctx).InvoiceLineItem.CreateBulk(builders...).Exec(ctx); err != nil {
     r.logger.Error("failed to add line items", "error", err)
     return ierr.WithError(err).WithHint("line item addition failed").Mark(ierr.ErrDatabase)
}
```

Creating new errors:
```go
if r.AmountDue.IsNegative() {
  return ierr.NewError("amount_due must be non-negative").
	  WithHint("amount due is negative").
	  WithReportableDetails(map[string]any{
		  "amount_due": r.AmountDue.String(),
	  }).Mark(ierr.ErrValidation)
}
```

`WithHint` - this is the public message that is returned back in the response
`WithReportableDetails` - this is the additional details that are returned back in the response

## Why `Mark` is important
This is important for our error handler middleware.  We use the function `ierr.HTTPStatusFromErr(err)` to get the http status, which is then passed to `c.JSON(status, response)` in our middleware to return the response.

```go
func HTTPStatusFromErr(err error) int {
	for e, status := range statusCodeMap {
		if errors.Is(err, e) {
			return status
		}
	}
	return http.StatusInternalServerError
}
```

Here the statusCodeMap is an exhaustive list of errors that we have defined:
```go
statusCodeMap = map[error]int{
		ErrHTTPClient:       http.StatusInternalServerError,
		ErrDatabase:         http.StatusInternalServerError,
		ErrNotFound:         http.StatusNotFound,
		ErrAlreadyExists:    http.StatusConflict,
		ErrVersionConflict:  http.StatusConflict,
		ErrValidation:       http.StatusBadRequest,
		ErrInvalidOperation: http.StatusBadRequest,
		ErrPermissionDenied: http.StatusForbidden,
		ErrSystem:           http.StatusInternalServerError,
}
```

## Sample Responses:

```json
{
    "success": false,
    "error": {
        "message": "request validation failed",
        "details": {
            "Currency": "Key: 'CreateInvoiceRequest.Currency' Error:Field validation for 'Currency' failed on the 'iso4217' tag",
            "InvoiceType": "Key: 'CreateInvoiceRequest.InvoiceType' Error:Field validation for 'InvoiceType' failed on the 'oneof' tag"
        }
    }
}
```

```json
{
    "success": false,
    "error": {
        "message": "invalid query filter",
        "details": {
            "limit": "must be between 1 and 1000"
        }
    }
}
```

----

> [!IMPORTANT]
> Refactor error handling using `cockroachdb/errors` and enhance invoice validation and error reporting across the codebase.
> 
>   - **Error Handling**:
>     - Introduces `ErrorBuilder` in `internal/errors/builder.go` for fluent error construction using `cockroachdb/errors`.
>     - Refactors error handling in `internal/errors/errors.go` to use `cockroachdb/errors` for error wrapping and hinting.
>     - Updates `internal/rest/middleware/errhandler.go` to use new error handling for consistent error responses.
>   - **Invoice Validation**:
>     - Enhances validation in `internal/api/dto/invoice.go` using `ErrorBuilder` for detailed error reporting.
>     - Updates `internal/domain/invoice/model.go` and `line_item.go` to use new error handling for validation errors.
>   - **Service and Repository Updates**:
>     - Refactors `internal/service/invoice.go` to use new error handling for invoice operations.
>     - Updates `internal/repository/ent/invoice.go` to handle database errors with `ErrorBuilder`.
>   - **Miscellaneous**:
>     - Changes `docker-compose.yml` to update port mappings for `temporal-ui` service.
>     - Adds new dependencies in `go.mod` and `go.sum` for `cockroachdb/errors` and related packages.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=flexprice%2Fflexprice&utm_source=github&utm_medium=referral)<sup> for bf6bdbb0c511018b74ea954eaeddf49a84403589. It will automatically update as commits are pushed.</sup>


internal/api/dto/invoice.go
```diff
@@ -17,8 +19,8 @@ type CreateInvoiceRequest struct {
 	CustomerID     string                         `json:"customer_id" validate:"required"`
 	SubscriptionID *string                        `json:"subscription_id,omitempty"`
 	IdempotencyKey *string                        `json:"idempotency_key"`
-	InvoiceType    types.InvoiceType              `json:"invoice_type" validate:"required"`
-	Currency       string                         `json:"currency" validate:"required"`
+	InvoiceType    types.InvoiceType              `json:"invoice_type" validate:"oneof=SUBSCRIPTION ONE_OFF CREDIT"`
+	Currency       string                         `json:"currency" validate:"required,iso4217"`
```
BRO3886: these validations were missing

instead of this we can also add [this](https://github.com/go-playground/validator/blob/master/_examples/struct-level/main.go#L81) - registering struct validations

internal/errors/errors.go
```diff
@@ -1,11 +1,14 @@
...
 import (
-	"errors"
 	"fmt"
+	"net/http"
+
+	"github.com/cockroachdb/errors"
 )
 
 // Common error types that can be used across the application
+// TODO: move to errors.New from cockroachdb/errors
```
BRO3886, line 11: when internal errors are no longer in use we can make these errors.new()

internal/rest/middleware/errhandler.go
```diff
@@ -0,0 +1,93 @@
...
+			if len(payload) > 9 && strings.HasPrefix(payload, "__json__:") {
+				jsonStr := payload[9:]
+				var jsonDetails map[string]any
+				if err := json.Unmarshal([]byte(jsonStr), &jsonDetails); err == nil {
+					for k, v := range jsonDetails {
+						details[k] = v // will overwrite any existing details?
+					}
+				}
+			}
+		}
```
nkmishra1997, line 85: Not able to understand this part very well and its intended use cases. When we already mark the safe and non dafe details while creating error why do we need to handle this ?
Plus things like `len(payload) > 9` make less sense to me

BRO3886: safety checks. If someone starts using `SafeDetails` without going through the builder, the format prefix might not have `__json__:`

```go
// WithReportableDetails adds structured details
func (b *ErrorBuilder) WithReportableDetails(details map[string]any) *ErrorBuilder {
	if len(details) == 0 {
		return b
	}
	marshaled, err := json.Marshal(details)
	if err != nil {
		return b
	}
	b.err = errors.WithSafeDetails(b.err, "__json__:%s", errors.Safe(string(marshaled)))
	return b
}
```

in the middleware we are looping through all safedetails in the error chain, and figuring out if the payload starts with `__json__:`. if it does, we are extracting the part after the `__json__:` prefix and unmarshaling that into a map. Since we are looping through multiple safedetails, one possible issue I see is keys being overwritten in the map. But this can be avoided once we ensure the payload is only created at the root level and after that errors are just passed around, or additional details are added (different keys)

internal/rest/middleware/errhandler.go
```diff
@@ -0,0 +1,93 @@
...
+	"encoding/json"
+	"strings"
+
+	"github.com/cockroachdb/errors"
+	ierr "github.com/flexprice/flexprice/internal/errors"
+	"github.com/gin-gonic/gin"
+)
+
+// ErrorResponse represents the standard error response structure
+type ErrorResponse struct {
```
nkmishra1997, line 13: todo: move this to dto/types to use for api docs

PR #154, opened 2025-04-05
feat(plan): add entitlement and subscription check logic for deletion

omkar273:
> [!IMPORTANT]
> Add checks for entitlements and subscriptions in `DeletePlan` to prevent deletion if associations exist.
> 
>   - **Behavior**:
>     - `DeletePlan` in `planService` now checks for associated entitlements and subscriptions before deletion, returning errors if any exist.
>     - Errors include hints and reportable details for better debugging.
>   - **Dependencies**:
>     - `planService` now requires `subscriptionRepo` for subscription checks.
>     - `onboardingService` and `subscriptionService` constructors updated to pass `subscriptionRepo` to `planService`.
>   - **Misc**:
>     - Minor formatting adjustments in `plan.go` and `subscription.go`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=flexprice%2Fflexprice&utm_source=github&utm_medium=referral)<sup> for 349ce66d5eb695b9962c8b32eae2a88df3701e14. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/internal/service/plan.go
+++ b/internal/service/plan.go
@@ -428,31 +428,22 @@ func (s *planService) UpdatePlan(ctx context.Context, id string, req dto.UpdateP
 
 func (s *planService) DeletePlan(ctx context.Context, id string) error {
 
-	entitlements, err := s.entitlementRepo.List(ctx, &types.EntitlementFilter{
-		QueryFilter: types.NewNoLimitQueryFilter(),
-		PlanIDs:     []string{id},
-	})
-
-	if err != nil {
-		return err
+	if id == "" {
+		return ierr.NewError("plan ID is required").
+			WithHint("Plan ID is required").
+			Mark(ierr.ErrValidation)
 	}
 
+	filters := types.NewNoLimitQueryFilter()
+	filters.Status = lo.ToPtr(types.StatusPublished)
+
 	subscriptions, err := s.subscriptionRepo.List(ctx, &types.SubscriptionFilter{
-		QueryFilter: types.NewNoLimitQueryFilter(),
+		QueryFilter: filters,
 		PlanID:      id,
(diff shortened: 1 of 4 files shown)


internal/service/plan.go
```diff
@@ -423,7 +427,44 @@ func (s *planService) UpdatePlan(ctx context.Context, id string, req dto.UpdateP
 }
 
 func (s *planService) DeletePlan(ctx context.Context, id string) error {
-	err := s.planRepo.Delete(ctx, id)
+
+	entitlements, err := s.entitlementRepo.List(ctx, &types.EntitlementFilter{
+		QueryFilter: types.NewNoLimitQueryFilter(),
+		PlanIDs:     []string{id},
+	})
```
nkmishra1997: also add status published filter here

internal/service/plan.go
```diff
@@ -423,7 +427,44 @@ func (s *planService) UpdatePlan(ctx context.Context, id string, req dto.UpdateP
...
+
+	if err != nil {
+		return err
+	}
+
+	subscriptions, err := s.subscriptionRepo.List(ctx, &types.SubscriptionFilter{
+		QueryFilter: types.NewNoLimitQueryFilter(),
+		PlanID:      id,
+	})
```
nkmishra1997: filter only active subscriptions

internal/service/plan.go
nkmishra1997: Also - I think we need to check only for subscriptions and not for entitlements.
Entitlements can not be deleted till they are linked to plans
and now if we do not allow deleting plans with entitlements - we end up in a deadlock.

PR #161, opened 2025-04-07
feat(customer): add checks before customer deletion

omkar273:


internal/service/customer.go
```diff
@@ -189,6 +195,43 @@ func (s *customerService) DeleteCustomer(ctx context.Context, id string) error {
 			Mark(ierr.ErrValidation)
 	}
 
+	subscriptionFilter := types.NewNoLimitSubscriptionFilter()
+	subscriptionFilter.CustomerID = id
+	subscriptions, err := s.subscriptionRepo.List(ctx, subscriptionFilter)
```
nkmishra1997, line 214: Subscription not cancelled check

internal/service/customer.go
```diff
@@ -189,6 +195,43 @@ func (s *customerService) DeleteCustomer(ctx context.Context, id string) error {
 			Mark(ierr.ErrValidation)
 	}
 
+	subscriptionFilter := types.NewNoLimitSubscriptionFilter()
```
nkmishra1997: Set Limit to 1 in these cases

PR #169, opened 2025-04-08
Add Pyroscope profiling integration and schema migration support

vaishnavi-cdf:
This PR adds the following:

- Integration of Pyroscope for continuous profiling
- Support for schema migration using Ent
- Flag `--dry-run` to show migration SQL without executing
- Logger setup for better debugging
- Safe database connection using DSN from config

@nk** Kindly review and let me know if any refinements are needed. 🚀


vaishnavi-cdf:
Done

PR #172, opened 2025-04-10
feat(invoice): s3 store for invoice pdfs

BRO3886:
> [!IMPORTANT]
> Adds functionality to store invoice PDFs in S3, including new S3 service, configuration updates, and error handling improvements.
> 
>   - **Behavior**:
>     - Adds `S3Service` interface and `s3ServiceImpl` in `service.go` to handle S3 operations: `UploadDocument`, `GetPresignedUrl`, `GetDocument`, and `Exists`.
>     - Supports only `DocumentTypeInvoice` for document operations.
>     - Presigned URL expiry defaults to 30 minutes if parsing fails.
>   - **Configuration**:
>     - Adds `S3Config` and `InvoiceBucketConfig` to `config.go` and `config.yaml` for S3 settings.
>     - Updates `main.go` to load AWS config and initialize S3 client and service.
>   - **Models**:
>     - Introduces `Document`, `DocumentKind`, and `DocumentType` in `model.go`.
>     - Adds `NewPdfDocument` constructor for PDF documents.
>   - **Dependencies**:
>     - Updates AWS SDK to `v1.36.3` in `go.mod` and `go.sum`.
>   - **Errors**:
>     - Adds `NewErrorf` function in `builder.go` for formatted error messages.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=flexprice%2Fflexprice&utm_source=github&utm_medium=referral)<sup> for 8f4fa84b0faeac9b15ae09931d4127fd7015a5c0. It will automatically update as commits are pushed.</sup>


internal/config/config.yaml
```diff
@@ -114,3 +114,15 @@ secrets:
...
+  enabled: true
+  region: "ap-south-1"
+  invoice:
+    bucket: "flexprice-invoices"
+    presign_expiry_duration: "1h"
+    key_prefix: ""
+  # can have multiple buckets like invoice, cdn, etc.
+  cdn:
+    bucket: "flexprice-cdn"
+    key_prefix: "cdn"
```
nkmishra1997: Lets remove if not using this in code

internal/config/config.yaml
```diff
@@ -114,3 +114,15 @@ secrets:
 billing:
   tenant_id: ""
   environment_id: ""
+
+s3:
+  enabled: true
+  region: "ap-south-1"
+  invoice:
+    bucket: "flexprice-invoices"
```
nkmishra1997, line 122: Please keep the default config to have `enabled: false` and this can be an additional config that people will set up in case they want to use.

nkmishra1997: Also add about all this in the SETUP.md as proper docs

nkmishra1997:
```suggestion
  enabled: false
  region: "us-east-1"
  invoice:
    bucket: "invoices-bucket"
```

nkmishra1997: And in the docs also mention that when you set this to true - you will have to set up a new S3 bucket and allow the infra where flexprice is running to have access to push to this S3 (ex in AWS ECS we have TaskRole where we can specify this)

nkmishra1997: From changes I can see we will need - PutObject, GetObject,HeadObject permissions

internal/service/invoice.go
```diff
@@ -792,6 +794,40 @@ func (s *invoiceService) performPaymentAttemptActions(ctx context.Context, inv *
 	return nil
 }
 
+func (s *invoiceService) GetInvoicePDFUrl(ctx context.Context, tenantId, id string) (string, error) {
```
nkmishra1997: I recommend to not add tenantID as input param but get it directly from context to prevent misuse

internal/api/v1/invoice.go
```diff
@@ -341,6 +341,7 @@ func (h *InvoiceHandler) AttemptPayment(c *gin.Context) {
 
 	c.JSON(http.StatusOK, gin.H{"message": "payment processed successfully"})
 }
+
 // GetInvoicePDF godoc
 // @Summary Get PDF for an invoice
 // @Description Retrieve the PDF document for a specific invoice by its ID
```
nkmishra1997, line 347: add the newly added param here as well so that is available when we generate docs

internal/config/config.yaml
BRO3886: yes was just mentioned to showcase how to create multiple buckets in config, removed

nkmishra1997 (approved):

PR #320, opened 2025-07-08
feat(tax): implement tax service

omkar273:


internal/domain/invoice/model.go
```diff
@@ -105,6 +105,8 @@ type Invoice struct {
 	// refunded_amount is the total sum of credit notes of type "refund".
 	// These are actual refunds issued to the customer.
 	RefundedAmount decimal.Decimal `json:"refunded_amount"`
+	// TotalTax holds the value of the "total_tax" field.
```
nkmishra1997: comment can be improved and should follow the way other comments are written

`total_tax` is the sum of all taxes combined at the invoice level.

internal/domain/tax/model.go
```diff
@@ -0,0 +1,70 @@
...
+	Description     string              `json:"description,omitempty"`
+	Code            string              `json:"code,omitempty"`
+	TaxRateStatus   types.TaxRateStatus `json:"tax_rate_status,omitempty"`
+	TaxRateType     types.TaxRateType   `json:"tax_rate_type,omitempty"`
+	Scope           types.TaxRateScope  `json:"scope,omitempty"`
+	PercentageValue *decimal.Decimal    `json:"percentage_value,omitempty"`
+	Currency        string              `json:"currency,omitempty"`
+	FixedValue      *decimal.Decimal    `json:"fixed_value,omitempty"`
+	ValidFrom       *time.Time          `json:"valid_from,omitempty"`
+	ValidTo         *time.Time          `json:"valid_to,omitempty"`
```
nkmishra1997: lets remove these as we do not have very good clarity on updating states and existing invoices/subscription

internal/domain/taxassociation/model.go
```diff
@@ -0,0 +1,61 @@
...
+	// Type of entity this tax rate applies to
+	EntityType types.TaxrateEntityType `json:"entity_type,omitempty"`
+	// ID of the entity this tax rate applies to
+	EntityID string `json:"entity_id,omitempty"`
+	// Priority for tax resolution (lower number = higher priority)
+	Priority int `json:"priority,omitempty"`
+	// Whether this tax should be automatically applied
+	AutoApply bool `json:"auto_apply,omitempty"`
+	// Currency
+	Currency string `json:"currency,omitempty"`
```
nkmishra1997, line 23: should we keep it at both association and rate level or only at association level

internal/domain/taxassociation/model.go
```diff
@@ -0,0 +1,61 @@
...
+	// Reference to the TaxRate entity
+	TaxRateID string `json:"tax_rate_id,omitempty"`
+	// Type of entity this tax rate applies to
+	EntityType types.TaxrateEntityType `json:"entity_type,omitempty"`
+	// ID of the entity this tax rate applies to
+	EntityID string `json:"entity_id,omitempty"`
+	// Priority for tax resolution (lower number = higher priority)
+	Priority int `json:"priority,omitempty"`
+	// Whether this tax should be automatically applied
+	AutoApply bool `json:"auto_apply,omitempty"`
```
nkmishra1997, line 21: should be default true?

ent/schema/taxrate.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Edges of the TaxRate.
+func (TaxRate) Edges() []ent.Edge {
+	return nil
+}
+
+// Indexes of the TaxRate.
+func (TaxRate) Indexes() []ent.Index {
+	return []ent.Index{
+		index.Fields("code", "tenant_id", "environment_id").
```
nkmishra1997: no need to be unique

ent/schema/taxrate.go
```diff
@@ -0,0 +1,113 @@
...
+			Unique().
+			StorageKey(Idx_code_tenant_id_environment_id).
+			Annotations(entsql.IndexWhere("(code IS NOT NULL AND code != '' and status = 'published')")),
+
+		index.Fields("code").
+			Annotations(entsql.IndexWhere("(code IS NOT NULL AND code != '')")),
```
nkmishra1997: no need

nkmishra1997: need indexes on tenant_id and environment_id

ent/schema/taxassociation.go
```diff
@@ -0,0 +1,123 @@
...
+
+		field.Int("priority").
+			Default(100).
+			SchemaType(map[string]string{
+				"postgres": "integer",
+			}).
+			Comment("Priority for tax resolution (lower number = higher priority)"),
+
+		field.Bool("auto_apply").
+			Default(false).
```
nkmishra1997: default true

ent/schema/taxassociation.go
```diff
@@ -0,0 +1,123 @@
...
+			StorageKey(Idx_entity_lookup_active),
+
+		// Tax rate reverse lookup: find entities using specific tax rate
+		index.Fields("tenant_id", "environment_id", "tax_rate_id").
+			StorageKey(Idx_tax_rate_id_tenant_id_environment_id),
+
+		// Auto-apply lookup for bulk operations
+		index.Fields("tenant_id", "environment_id", "auto_apply", "entity_type").
+			StorageKey(Idx_auto_apply_lookup),
+
```
nkmishra1997: no need

ent/schema/taxassociation.go
```diff
@@ -0,0 +1,123 @@
...
+// Indexes of the TaxAssociation.
+func (TaxAssociation) Indexes() []ent.Index {
+	return []ent.Index{
+		// Primary lookup: find tax assignments for entity
+		index.Fields("tenant_id", "environment_id", "entity_type", "entity_id", "status").
+			StorageKey(Idx_entity_lookup_active),
+
+		// Tax rate reverse lookup: find entities using specific tax rate
+		index.Fields("tenant_id", "environment_id", "tax_rate_id").
+			StorageKey(Idx_tax_rate_id_tenant_id_environment_id),
```
nkmishra1997, line 106: remove

ent/schema/taxassociation.go
```diff
@@ -0,0 +1,123 @@
...
+		// Unique constraint: prevent duplicate assignments per entity
+		index.Fields("tenant_id", "environment_id", "entity_type", "entity_id", "tax_rate_id").
+			StorageKey(Unique_entity_tax_mapping).
+			Unique(),
+
+		// Currency-based lookup
+
+		index.Fields("tenant_id", "environment_id", "currency", "entity_type").
+			StorageKey(Idx_currency_entity_lookup),
+	}
```
nkmishra1997: remove and move the currency also in the main index

ent/schema/tax_applied.go
```diff
@@ -0,0 +1,135 @@
...
+	return []ent.Index{
+		// Primary lookup: find tax applications for entity and tax rate
+		index.Fields("tenant_id", "environment_id", "entity_type", "entity_id", "tax_rate_id").
+			Unique().
+			StorageKey(Idx_entity_tax_rate_id),
+
+		// Secondary lookup: find tax applications for entity and tax association
+		index.Fields("tenant_id", "environment_id", "entity_type", "entity_id").
+			StorageKey(Idx_entity_tax_association_lookup),
+	}
```
nkmishra1997: the larger index should be enough but need to confirm

internal/api/dto/invoice.go
```diff
@@ -76,6 +76,9 @@ type CreateInvoiceRequest struct {
 
 	// environment_id is the unique identifier of the environment this invoice belongs to
 	EnvironmentID string `json:"environment_id,omitempty"`
```
nkmishra1997: why adding env id as request input param? this should come from the ctx.

internal/api/dto/customer.go
```diff
@@ -4,45 +4,109 @@ import (
...
+	AddressState *string `json:"address_state" validate:"omitempty,max=100"`
+
+	// address_postal_code is the updated postal code with maximum 20 characters
+	AddressPostalCode *string `json:"address_postal_code" validate:"omitempty,max=20"`
+
+	// address_country is the updated two-letter ISO 3166-1 alpha-2 country code
+	AddressCountry *string `json:"address_country" validate:"omitempty,len=2,iso3166_1_alpha2"`
+
+	// metadata contains updated key-value pairs that will replace existing metadata
+	Metadata map[string]string `json:"metadata,omitempty"`
```
nkmishra1997, line 80: not letting updation of associated tax rate?

nkmishra1997: In case this is concious call just document the reason here

internal/api/dto/invoice.go
```diff
@@ -74,8 +74,8 @@ type CreateInvoiceRequest struct {
 	// metadata contains additional custom key-value pairs for storing extra information
 	Metadata types.Metadata `json:"metadata,omitempty"`
 
-	// environment_id is the unique identifier of the environment this invoice belongs to
-	EnvironmentID string `json:"environment_id,omitempty"`
+	// tax_rate_overrides is the tax rate overrides to be applied to the invoice
+	TaxRateOverrides []*TaxRateOverride `json:"tax_rate_overrides,omitempty"`
```
nkmishra1997, line 78: While creating invoice, we might want to support providing taxes directly with the invoice

Can start with only the existing taxes for now but eventually we need to support all cases:
1. Taxes directly coming on invoices from APIs and we always create them in system just that they might or might now get created as one off cases.
2. Taxes flowing in from integrations like stripe tax / avalara
3. Pre configured taxes being auto applied

internal/api/router.go
```diff
@@ -265,6 +266,25 @@ func NewRouter(handlers Handlers, cfg *config.Configuration, logger *logger.Logg
...
+		}
+
+		tax := v1Private.Group("/taxassociations")
+		{
+			tax.POST("", handlers.Tax.CreateTaxAssociation)
+			tax.GET("", handlers.Tax.ListTaxAssociations)
+			tax.GET("/:id", handlers.Tax.GetTaxAssociation)
+			tax.PUT("/:id", handlers.Tax.UpdateTaxAssociation)
+			tax.DELETE("/:id", handlers.Tax.DeleteTaxAssociation)
+		}
```
nkmishra1997: From what I am understanding - we are explicitly exposing APIs to create tax associations and also adding them in the Create/Update of the eligible entities i.e customer, subscription and invoices? 

Need to be super careful in testing both flows and be mindful here

internal/api/router.go
```diff
@@ -265,6 +266,25 @@ func NewRouter(handlers Handlers, cfg *config.Configuration, logger *logger.Logg
 			tasks.POST("/:id/process", handlers.Task.ProcessTask)
 		}
 
+		// Tax rate routes
+		taxRates := v1Private.Group("/taxrates")
```
nkmishra1997:
```suggestion
		taxRates := v1Private.Group("/taxes/rates")
```

internal/api/router.go
```diff
@@ -265,6 +266,25 @@ func NewRouter(handlers Handlers, cfg *config.Configuration, logger *logger.Logg
...
+		{
+			taxRates.POST("", handlers.Tax.CreateTaxRate)
+			taxRates.GET("", handlers.Tax.ListTaxRates)
+			taxRates.GET("/:id", handlers.Tax.GetTaxRate)
+			taxRates.PUT("/:id", handlers.Tax.UpdateTaxRate)
+			taxRates.DELETE("/:id", handlers.Tax.DeleteTaxRate)
+		}
+
+		tax := v1Private.Group("/taxassociations")
```
nkmishra1997:
```suggestion
		tax := v1Private.Group("/taxes/associations")
```

internal/repository/factory.go
```diff
@@ -21,6 +21,9 @@ import (
 	"github.com/flexprice/flexprice/internal/domain/secret"
 	"github.com/flexprice/flexprice/internal/domain/subscription"
 	"github.com/flexprice/flexprice/internal/domain/task"
+	taxrate "github.com/flexprice/flexprice/internal/domain/tax"
+	taxapplied "github.com/flexprice/flexprice/internal/domain/taxapplied"
+	taxconfig "github.com/flexprice/flexprice/internal/domain/taxassociation"
```
nkmishra1997: can call this taxAssociation directly no need of translating it as taxconfig

internal/service/billing.go
```diff
@@ -712,7 +712,6 @@ func (s *billingService) CreateInvoiceRequestForCharges(
 		PeriodStart:    &periodStart,
 		PeriodEnd:      &periodEnd,
 		BillingReason:  types.InvoiceBillingReasonSubscriptionCycle,
```
nkmishra1997, line 730: just want to be sure that we are removing environment_id from create invoice request I hope it was not added there due to some reason (maybe cron? just need to verify that and ideally it should not be there)

internal/service/customer.go
```diff
@@ -44,10 +44,49 @@ func (s *customerService) CreateCustomer(ctx context.Context, req dto.CreateCust
...
+			}
+
+			err = taxService.LinkTaxRatesToEntity(txCtx, dto.LinkTaxRateToEntityRequest{
+				EntityType:              types.TaxrateEntityTypeCustomer,
+				EntityID:                cust.ID,
+				ExistingTaxAssociations: tenantTaxAssociations.Items,
+			})
+			if err != nil {
+				return err
+			}
```
nkmishra1997, line 86: I don't think we should pick and auto associate tenant level tax rates to each customer.  This should happen in the resolution layer direclty to pick taxes at all level and then pick the right one basis priority

omkar273: This is done to fulfil the usecase where, We need to autoapply taxes to new customers which are configured to autoapply at tenant level

internal/api/router.go
omkar273: This is done to fulfil the use case where the user wants to
link or unlink a specific association to a customer or subscription.
So, instead of calling the update API, it is efficient to use the tax associations APIs, which will be faster and efficient

internal/api/dto/invoice.go
omkar273: For that, I have planned it in a way 
I will add field taxes, which will be an array of create tax rate payload
such that at a single time, a user can either send tax overrides or taxes 
In this way, we will be sure whether the invoice taxes will be internal or external

internal/service/billing.go
omkar273: I checked its usage it is not being used in any cron it's being handled in ToInvoice method 
```go
if r.EnvironmentID != "" {
     nv.EnvironmentID = r.EnvironmentID
} else {
     inv.EnvironmentID = types.GetEnvironmentID(ctx)
}
```

PR #325, opened 2025-07-11
feat(price): implement fiat-based custom pricing units and integrate with pricing workflow

subratsahilgupta:
**This PR includes:**

> 
> Custom Pricing Unit workflow
> 
> Integration of custom pricing units into the pricing flow

**Custom Pricing Unit Validation Rules**

> 1. Custom pricing unit code must:
>    - Be 3 characters long
>    - Be unique within tenant's environment
>    - Not conflict with existing fiat currency codes
> 2. Base currency must be from supported fiat currencies
> 3. Conversion rate must be positive non-zero decimal
> 4. Precision must be between 0-8
> 5. Create Validations:
>       Tenant ID required
>       Environment ID required
>       Code must be unique for published units
>       Conversion rate required
> 6. Update Status Transitions:
>       Published -> Archived
>       Archived -> Published (with code uniqueness check)
>       Archived -> Deleted
>       No other transitions allowed
>       Can't transition to same status
> 7. Update Field Validations:
>       Name, Symbol, and Precision are optional
>       Status transitions are validated
>       Code uniqueness checked when going back to Published

**Pricing Flow Logic**

> When creating/updating a price:
> 1. If custom_price_unit is provided in request:
>    - Validate custom price unit exists and is active
>    - Get conversion rate and precision from custom price unit
>    - Store base currency amount (amount * conversion_rate)
>    - Store custom price unit amount (original amount)
>    - Store custom price unit code
> 2. If no custom_price_unit:
>    - Regular price flow continues
> When returning price:
> 1. If custom_price_unit exists:
>    - Return both base currency amount and custom price unit amount
>    - Format display amounts for both currencies
>    - Include conversion rate and custom unit details
> 2. If no custom_price_unit:
>    - Return regular price response

✅ Tested and validated locally via API calls.

_Upcoming Enhancements-_ 

> Support in Subscription Creation/View
> Support in Invoice Creation/View
> 
> Unit tests for all enhanced service layer components

 

----

> [!IMPORTANT]
> Implement fiat-based custom pricing units and integrate them into the pricing workflow with validation and API support.
> 
>   - **Custom Pricing Unit Workflow**:
>     - Implement fiat-based custom pricing units in `custom_pricing_unit.go`.
>     - Add validation rules for custom pricing units in `custom_pricing_unit.go`.
>     - Integrate custom pricing units into pricing flow in `price.go`.
>   - **API and Handlers**:
>     - Add `CustomPricingUnitHandler` in `custom_pricing_unit.go` for CRUD operations.
>     - Update `router.go` to include routes for custom pricing units.
>   - **Database and Models**:
>     - Add `CustomPricingUnit` schema in `custompricingunit.go`.
>     - Update `schema.go` and `client.go` to include `CustomPricingUnit`.
>     - Modify `price.go` to support custom pricing unit fields.
>   - **Services**:
>     - Implement `CustomPricingUnitService` in `custom_pricing_unit.go`.
>     - Update `PriceService` to handle custom pricing units in `price.go`.
>   - **Miscellaneous**:
>     - Update `price_create.go`, `price_update.go`, and `price_query.go` for custom pricing unit integration.
>     - Add tests and documentation for new features.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=flexprice%2Fflexprice&utm_source=github&utm_medium=referral)<sup> for 0c555761bcf7b8e1c0b116d794799bdd1d414e96. You can [customize](https://app.ellipsis.dev/flexprice/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


ent/schema/price.go
```diff
@@ -46,6 +46,28 @@ func (Price) Fields() []ent.Field {
...
+		field.Float("price_unit_amount").
+			SchemaType(map[string]string{
+				"postgres": "numeric(25,15)",
+			}).
+			Optional(),
+		field.String("display_price_unit_amount").
+			SchemaType(map[string]string{
+				"postgres": "varchar(255)",
+			}).
+			Optional(),
```
nkmishra1997, line 63: Let's remove this

nkmishra1997: all customer_pricing_unit can have same prefix else it is confusing
weather the conversion rate is for custom pricing unit or something

can try:
price_unit_*

price_unit_id
price_unit_code
price_unit_amount
price_unit_conversion_factor

can just keep these fields maybe and remove others and pull them from pricing_unit table and enrich in the DTO if required. Something like:
```json
.....existing fields
"custom_pricing_unit": {
entire custom pricing response DTO
}
.....existing fields
```


can follow this exact pattern for subscription line items and invoice line items as well
in the LIST APIs of these entities can add a expand param `pricing_unit` which, when passed will expand the params for all the results.

internal/api/dto/custom_pricing_unit.go
```diff
@@ -0,0 +1,50 @@
...
+	ConversionRate *decimal.Decimal `json:"conversion_rate" validate:"required,gt=0"`
+	Precision      int              `json:"precision" validate:"gte=0,lte=8"`
+}
+
+// UpdateCustomPricingUnitRequest represents the request to update an existing custom pricing unit
+type UpdateCustomPricingUnitRequest struct {
+	Name      string       `json:"name,omitempty" validate:"omitempty"`
+	Symbol    string       `json:"symbol,omitempty" validate:"omitempty,max=10"`
+	Precision int          `json:"precision,omitempty" validate:"omitempty,gte=0,lte=8"`
+	Status    types.Status `json:"status" validate:"omitempty,oneof=published archived deleted"`
```
nkmishra1997, line 24: can remove status
also we need to avoid oneof tag in the validate tags as it leads to issue in SDK generation. Hence we have implemented type specific validations and use them for such enum fields

internal/api/dto/custom_pricing_unit.go
```diff
@@ -0,0 +1,50 @@
...
+	BaseCurrency   string          `json:"base_currency"`
+	ConversionRate decimal.Decimal `json:"conversion_rate"`
+	Precision      int             `json:"precision"`
+	Status         types.Status    `json:"status"`
+	CreatedAt      string          `json:"created_at"`
+	UpdatedAt      string          `json:"updated_at"`
+}
+
+// ListCustomPricingUnitsResponse represents the paginated response for listing custom pricing units
+type ListCustomPricingUnitsResponse struct {
```
nkmishra1997, line 42: pls use the template response of List entity ex check
```go
// ListPricesResponse represents the response for listing prices
type ListPricesResponse = types.ListResponse[*PriceResponse]
```

internal/api/dto/price.go
```diff
@@ -14,6 +14,7 @@ import (
 type CreatePriceRequest struct {
 	Amount             string                   `json:"amount"`
 	Currency           string                   `json:"currency" validate:"required,len=3"`
+	PriceUnit          string                   `json:"price_unit,omitempty"`
```
nkmishra1997, line 17: if we are doing just this, then essentially the caller is responsible to make sure that they convert the `amount` field to the desired value?

nkmishra1997: can follow same patter here as well maybe:
price_unit_code
price_unit_amount

but my only concern with that is that we need to then make amount and currency optional which is like making 99% case optional for 1% case.

nkmishra1997: Let me come back on this

internal/api/router.go
```diff
@@ -132,6 +133,16 @@ func NewRouter(handlers Handlers, cfg *config.Configuration, logger *logger.Logg
 			price.DELETE("/:id", handlers.Price.DeletePrice)
 		}
 
+		customPricingUnit := v1Private.Group("/pricing_units")
```
nkmishra1997, line 136:
```suggestion
		customPricingUnit := v1Private.Group("/pricing/units")
```

PR #353, opened 2025-07-30
Feat/stripe sync v5

pratham-ez:


internal/domain/connection/repository.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+
+	"github.com/flexprice/flexprice/internal/types"
+)
+
+// Repository defines the interface for connection data operations
+type Repository interface {
+	Create(ctx context.Context, connection *Connection) error
+	Get(ctx context.Context, id string) (*Connection, error)
+	GetByEnvironmentAndProvider(ctx context.Context, environmentID string, provider types.SecretProvider) (*Connection, error)
```
nkmishra1997, line 13: can remove this or even if we keep we can call List method internally

internal/domain/entityintegrationmapping/model.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// ValidateProviderType validates the provider type
+func ValidateProviderType(providerType string) bool {
+	validProviders := map[string]bool{
+		"stripe":   true,
+		"razorpay": true,
+		"paypal":   true,
+	}
+	return validProviders[providerType]
```
nkmishra1997, line 88: create enums and move to types/ folder. Refer `type SubscriptionStatus string` as example

internal/domain/entityintegrationmapping/model.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	if !ValidateProviderType(m.ProviderType) {
+		return ierr.NewError("invalid provider_type").
+			WithHint("Provider type must be one of: stripe, razorpay, paypal").
+			Mark(ierr.ErrValidation)
+	}
+
+	if m.ProviderEntityID == "" {
+		return ierr.NewError("provider_entity_id is required").
```
nkmishra1997, line 124: we can ignore these checks and keep them in DTO layer only. This is to be user for more logical checks than static ones

internal/domain/entityintegrationmapping/repository.go
```diff
@@ -0,0 +1,24 @@
...
+	Count(ctx context.Context, filter *types.EntityIntegrationMappingFilter) (int, error)
+	ListAll(ctx context.Context, filter *types.EntityIntegrationMappingFilter) ([]*EntityIntegrationMapping, error)
+	Update(ctx context.Context, mapping *EntityIntegrationMapping) error
+	Delete(ctx context.Context, mapping *EntityIntegrationMapping) error
+
+	// Provider-specific queries
+	GetByEntityAndProvider(ctx context.Context, entityID, entityType, providerType string) (*EntityIntegrationMapping, error)
+	GetByProviderEntity(ctx context.Context, providerType, providerEntityID string) (*EntityIntegrationMapping, error)
+	ListByEntity(ctx context.Context, entityID, entityType string) ([]*EntityIntegrationMapping, error)
+	ListByProvider(ctx context.Context, providerType string) ([]*EntityIntegrationMapping, error)
```
nkmishra1997, line 23: use list methods when implementing these in ent repo layer

PR #466, opened 2025-08-30
Feat/upgrade downgrade

hiteshshimpi-55:


internal/api/dto/subscription_change.go
```diff
@@ -0,0 +1,276 @@
...
+	return nil
+}
+
+// SubscriptionChangePreviewRequest represents a request to preview subscription changes
+// @Description Request object for previewing the effects of a subscription plan change
+type SubscriptionChangePreviewRequest struct {
+	SubscriptionChangeRequest
+
+	// preview_date is the date to calculate the preview for (optional, defaults to effective_date or now)
+	PreviewDate *time.Time `json:"preview_date,omitempty"`
```
nkmishra1997: rethink if we need this - can simplify this only to use effective_date for preview case as well

internal/api/dto/subscription.go
```diff
@@ -86,6 +96,72 @@ type UpdateSubscriptionRequest struct {
 	CancelAtPeriodEnd bool                     `json:"cancel_at_period_end,omitempty"`
 }
```
nkmishra1997, line 97: later: deprecate this

internal/service/subscription.go
```diff
@@ -3,25 +3,29 @@ package service
...
 	webhookDto "github.com/flexprice/flexprice/internal/webhook/dto"
 	"github.com/samber/lo"
 	"github.com/shopspring/decimal"
+	"go.uber.org/zap"
 )
 
 type SubscriptionService interface {
 	CreateSubscription(ctx context.Context, req dto.CreateSubscriptionRequest) (*dto.SubscriptionResponse, error)
 	GetSubscription(ctx context.Context, id string) (*dto.SubscriptionResponse, error)
 	CancelSubscription(ctx context.Context, id string, cancelAtPeriodEnd bool) error
```
nkmishra1997: todo deprecate

PR #484, opened 2025-09-04
feat(subscription): add subscription renewal due webhook alert workflow

subratsahilgupta:


internal/repository/ent/subscription.go
```diff
@@ -330,6 +330,40 @@ func (r *subscriptionRepository) List(ctx context.Context, filter *types.Subscri
 	return result, nil
 }
 
+// ListActiveSubscriptionsDueForRenewal retrieves all active subscriptions that are due for renewal in 24 hours
+func (r *subscriptionRepository) ListSubscriptionsDueForRenewal(ctx context.Context) ([]*domainSub.Subscription, error) {
```
nkmishra1997, line 334: For future - can have limits and offset in this to avoid fetching 1000s of records

nkmishra1997 (approved):

PR #487, opened 2025-09-04
feat(temporal): migrate plan sync to temporal

omkar273:


internal/temporal/service.go
```diff
@@ -59,34 +60,33 @@ func (s *Service) StartBillingWorkflow(ctx context.Context, input models.Billing
...
 	workflowOptions := client.StartWorkflowOptions{
 		ID:        workflowID,
 		TaskQueue: s.cfg.TaskQueue,
 	}
 
-	planService := service.NewPlanService(s.ServiceParams)
-	we, err := s.client.Client.ExecuteWorkflow(ctx, workflowOptions, "PriceSyncWorkflow", models.PriceSyncWorkflowInput{
-		PlanID:       planID,
-		PriceService: planService,
+	we, err := s.client.Client.ExecuteWorkflow(ctx, workflowOptions, string(types.PriceSyncWorkflow), models.PriceSyncWorkflowInput{
```
nkmishra1997: need to check how this workflow is getting executed weather it is async or sync
and we ideally need to make this always async that we just trigger temporal and we get the run id
and from that point we just query the status of that run id (either manually or via APIs) to know final status

internal/temporal/workflows/price_sync_workflow.go
```diff
@@ -1,36 +1,53 @@
...
+		PlanID:        in.PlanID,
+		TenantID:      in.TenantID,
+		EnvironmentID: in.EnvironmentID,
+	}
 
-	activityOptions := workflow.ActivityOptions{
-		StartToCloseTimeout: time.Minute * 5,
-		RetryPolicy: &temporalsdk.RetryPolicy{
+	ao := workflow.ActivityOptions{
+		StartToCloseTimeout: 5 * time.Minute,
```
nkmishra1997: 5 mins might be less for this run - should give 30 mins

internal/types/activity.go
```diff
@@ -8,12 +8,22 @@ import (
 type ActivityType string
 
 const (
-	// Activity Types
+	// Activity Types - must match the method names in activity structs
 	ActivitySyncPlanPrices ActivityType = "SyncPlanPrices"
```
nkmishra1997: NIT: slightly generic as activities and workflows sound generic here and not temporal specific

internal/api/dto/subscription.go
```diff
@@ -84,7 +84,7 @@ type CreateSubscriptionRequest struct {
 	BillingAnchor *time.Time `json:"-"`
 
 	// Workflow
-	Workflow *types.WorkflowType `json:"-"`
+	Workflow *types.TemporalWorkflowType `json:"-"`
```
nkmishra1997, line 87: @hiteshshimpi-55 what is the use case for this?

nkmishra1997 (approved):

PR #498, opened 2025-09-09
refactor(temporal): migrate to clean architecture

omkar273:
> [!IMPORTANT]
> Refactor Temporal integration to clean architecture with event-driven processing, restructuring client, service, and worker components, and introducing new interfaces and models.
> 
>   - **Architecture**:
>     - Refactor Temporal integration to follow clean architecture principles.
>     - Introduce event-driven architecture to decouple API layer from Temporal.
>     - Restructure Temporal client, service, and worker components.
>   - **Components**:
>     - New interfaces for `TemporalClient`, `TemporalService`, and `TemporalWorker`.
>     - Introduce `models` for workflow inputs/outputs and options.
>     - Implement `worker` package for managing Temporal workers.
>   - **Functionality**:
>     - Add event-driven processing for tasks using `EventBus` and `TaskEventHandler`.
>     - Implement `TaskProcessingWorkflow` and `TaskActivities` for task processing.
>     - Support multiple file providers in `FileProviderRegistry` for file processing.
>   - **Misc**:
>     - Update `types` for context management and task queue definitions.
>     - Add utility functions for UUID generation with workflow context.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=flexprice%2Fflexprice&utm_source=github&utm_medium=referral)<sup> for 9a91e01d23aa6fb7c8eb2fe230cdad7276a7c43b. You can [customize](https://app.ellipsis.dev/flexprice/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>


cmd/server/main.go
```diff
@@ -360,13 +396,16 @@ func startServer(
...
 		startMessageRouter(lc, router, webhookService, onboardingService, log)
 		startPostProcessingConsumer(lc, router, eventPostProcessingSvc, cfg, log)
-		startTemporalWorker(lc, temporalClient, &cfg.Temporal, params)
+		startTemporalWorker(lc, temporalService, params)
 	case types.ModeAPI:
 		startAPIServer(lc, r, cfg, log)
 		startMessageRouter(lc, router, webhookService, onboardingService, log)
+		startTemporalWorker(lc, temporalService, params)
+		startConsumer(lc, consumer, eventRepo, cfg, log, sentryService, eventPostProcessingSvc)
+		startPostProcessingConsumer(lc, router, eventPostProcessingSvc, cfg, log)
```
nkmishra1997: why adding all this in the API mode? should be just the API here not other things

internal/types/temporal.go
```diff
@@ -8,68 +8,49 @@ import (
...
-	TemporalActivityFetchData            TemporalActivityType = "FetchData"
-	TemporalActivityCalculate            TemporalActivityType = "Calculate"
-	TemporalActivitySubscriptionChange   TemporalActivityType = "SubscriptionChange"
-	TemporalActivitySubscriptionCreation TemporalActivityType = "SubscriptionCreation"
+	// Task Queues - logical groupings to limit worker count
+	TemporalTaskQueueBilling      TemporalTaskQueue = "billing"
+	TemporalTaskQueueSubscription TemporalTaskQueue = "subscription"
+	TemporalTaskQueueTask         TemporalTaskQueue = "task"
+	TemporalTaskQueuePrice        TemporalTaskQueue = "price"
+	TemporalTaskQueueNotification TemporalTaskQueue = "notification"
```
nkmishra1997: i think module by module level will be too much we can go use cases by usecase:
1. notifications
2. tasks
3. alerts/webhooks like this

nkmishra1997: For later - for now anyways these are not getting used,

internal/temporal/registration.go
```diff
@@ -1,28 +1,81 @@
...
+
+// buildWorkerConfig creates a worker configuration for a specific task queue
+func buildWorkerConfig(taskQueue types.TemporalTaskQueue, planActivities *activities.PlanActivities, taskActivities *activities.TaskActivities) WorkerConfig {
+	workflowsList := []interface{}{}
+	activitiesList := []interface{}{}
+
+	switch taskQueue {
+	case types.TemporalTaskQueueTask:
+		workflowsList = append(workflowsList, workflows.TaskProcessingWorkflow)
+		activitiesList = append(activitiesList, taskActivities.ProcessTask)
```
nkmishra1997, line 48: in future there can be common activities list as well right? say send webhook can be an activity which is being used by myltiple workflows and will need to be explicitly defined here at a queue level atleast. Just confirming the behaviour here.

cmd/server/main.go
omkar273: Missed it

internal/types/temporal.go
omkar273: Noted

internal/temporal/activities/task_activities.go
```diff
@@ -0,0 +1,118 @@
...
+		ProcessedRecords:  task.ProcessedRecords,
+		SuccessfulRecords: task.SuccessfulRecords,
+		FailedRecords:     task.FailedRecords,
+		ErrorSummary:      task.ErrorSummary,
+		Metadata:          task.Metadata,
+	}, nil
+}
+
+// sendHeartbeats sends periodic heartbeats to Temporal
+func (a *TaskActivities) sendHeartbeats(ctx context.Context, taskID string) {
```
nkmishra1997: remove if not being used anywhere as it will increase number of goroutnies

internal/temporal/client/helper.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/flexprice/flexprice/internal/types"
+)
+
+// WaitForHealthy waits for the client to become healthy with a timeout
+func WaitForHealthy(ctx context.Context, client TemporalClient, timeout time.Duration) error {
```
nkmishra1997: remove is not being used?

internal/service/event_driven_processing.md
```diff
@@ -0,0 +1,496 @@
+# Event-Driven File Processing Architecture
+
```
nkmishra1997: lets move this to prd

internal/service/technical_deep_dive_diagram.md
```diff
@@ -0,0 +1,195 @@
+# Temporal Architecture Redesign
+
+## Current Issues
+
```
nkmishra1997, line 4: move to prds

internal/temporal/workflows/task_workflow.go
```diff
@@ -0,0 +1,79 @@
...
+
+	logger := workflow.GetLogger(ctx)
+	logger.Info("Starting task processing workflow", "task_id", input.TaskID)
+
+	// Define activity options with extended timeouts for large file processing
+	ao := workflow.ActivityOptions{
+		StartToCloseTimeout:    time.Hour * 4,   // Extended to 4 hours for very large files
+		ScheduleToCloseTimeout: time.Hour * 4,   // Total time from scheduling to completion
+		ScheduleToStartTimeout: time.Minute * 5, // Time to wait for activity to start
+		HeartbeatTimeout:       time.Minute * 2, // Heartbeat timeout for long-running activities
```
nkmishra1997: what happens if we do not pass the heartbeat? Any impact on the job?

internal/temporal/workflows/task_workflow.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	logger := workflow.GetLogger(ctx)
+	logger.Info("Starting task processing workflow", "task_id", input.TaskID)
+
+	// Define activity options with extended timeouts for large file processing
+	ao := workflow.ActivityOptions{
+		StartToCloseTimeout:    time.Hour * 4,   // Extended to 4 hours for very large files
+		ScheduleToCloseTimeout: time.Hour * 4,   // Total time from scheduling to completion
+		ScheduleToStartTimeout: time.Minute * 5, // Time to wait for activity to start
```
nkmishra1997: Do we really need this? Let temporal take as much as time to start the activity?

internal/temporal/service/service.go
```diff
@@ -0,0 +1,385 @@
...
+		return nil, errors.WithError(err).
+			WithHint("Invalid workflow type provided").
+			Mark(errors.ErrValidation)
+	}
+
+	return s.client.StartWorkflow(ctx, options, workflow, args...)
+}
+
+// SignalWorkflow implements TemporalService
+func (s *temporalService) SignalWorkflow(ctx context.Context, workflowID, runID, signalName string, arg interface{}) error {
```
nkmishra1997, line 79: what is the usecase for this method?

internal/service/file_processor.go
```diff
@@ -0,0 +1,140 @@
...
+	return &FileProcessor{
+		StreamingProcessor: NewStreamingProcessor(client, logger),
+		ProviderRegistry:   NewFileProviderRegistry(),
+		CSVProcessor:       NewCSVProcessor(logger),
+		JSONProcessor:      NewJSONProcessor(logger),
+	}
+}
+
+// DownloadFile downloads a file and returns the full content (for regular processing)
+func (fp *FileProcessor) DownloadFile(ctx context.Context, t *task.Task) ([]byte, error) {
```
nkmishra1997, line 66: When we download the file, does this method save the file on the disk? I can see that we are downloading the file using GET request and returning the body as raw bytes

In case this method downlaods the file on disk, we need to delete it later else the machine storage will be full.

In case it does not save the file on the disk but keeps it in memory, we will have to later optimise this as in case of large files it will kill the tasks by going Out Of Memory

internal/service/streaming_processor.go
```diff
@@ -0,0 +1,571 @@
...
+	maxFailures  int
+	failureCount int
+	lastFailure  time.Time
+	resetTimeout time.Duration
+	state        CircuitBreakerState
+	mutex        sync.RWMutex
+}
+
+// NewCircuitBreaker creates a new circuit breaker
+func NewCircuitBreaker(maxFailures int, resetTimeout time.Duration) *CircuitBreaker {
```
nkmishra1997: I recommend not to implement this custom written circuit breaker and rather in the first version -  just simplify without a circuit breaker or else use any existing library for circuit breaking.

We need to be sure of closing and opening logic of circuit breaking

internal/service/streaming_processor.go
```diff
@@ -0,0 +1,571 @@
...
+			}
+			continue
+		}
+
+		chunk = append(chunk, record)
+
+		// Process chunk when it reaches the configured size
+		if len(chunk) >= config.ChunkSize {
+			// Check circuit breaker
+			if !sp.CircuitBreaker.CanExecute() {
```
nkmishra1997: I can see logic for circuit breaker going into 5 failures to mark it as open and a single success to make it go closed but how does it recover to closed after the configured threshold of failure?
And what happens to the failed chunks during the open state?

internal/api/v1/task.go
```diff
@@ -100,108 +118,124 @@ func (h *TaskHandler) GetTask(c *gin.Context) {
...
+// @Accept json
+// @Produce json
+// @Security ApiKeyAuth
+// @Param id path string true "Task ID"
+// @Success 200 {object} gin.H
+// @Failure 400 {object} ierr.ErrorResponse
+// @Failure 404 {object} ierr.ErrorResponse
+// @Failure 500 {object} ierr.ErrorResponse
+// @Router /tasks/{id}/process [post]
+func (h *TaskHandler) ProcessTask(c *gin.Context) {
```
nkmishra1997: when is this method called? What if the task is already in processing state?

nkmishra1997: If this is not being used anywhere and now since we are using temporal based workflows - does it makes sense to have a process method which will start executing the workflow on the api server ?

internal/service/file_processor.go
omkar273: I can see that it does NOT save the file to disk. The method downloads the file content into memory and returns it as raw bytes ([]byte).

internal/api/v1/task.go
```diff
@@ -7,85 +7,103 @@ import (
...
 	if err != nil {
-		h.logger.Error("failed to create task", "error", err)
 		c.Error(err)
 		return
 	}
 
-	c.JSON(http.StatusCreated, task)
+	// Start the temporal workflow for async processing using the unified method
+	workflowRun, err := h.temporalService.ExecuteWorkflow(c.Request.Context(), types.TemporalTaskProcessingWorkflow, resp.ID)
+
```
nkmishra1997, line 63: NIT: ideally this should be part of task service not handler

internal/api/v1/task.go
```diff
@@ -7,85 +7,103 @@ import (
...
+	if err != nil {
+		h.log.Error("failed to start temporal workflow", "error", err, "task_id", resp.ID)
+		c.Error(err)
+		return
+	}
+
+	c.JSON(http.StatusOK, models.TemporalWorkflowResult{
+		Message:    "task processing workflow started successfully",
+		WorkflowID: workflowRun.GetID(),
+		RunID:      workflowRun.GetRunID(),
```
nkmishra1997, line 73: Ideally workflow and run id should be stored in task metadata or i think even better to create a column.

Also need to implement a retry api as well so that we can ask temporal to retry the task and update the run id after retry

Also need to check if we can cancel a task? I think temporal allows terminating a task so maybe we can use that as well

internal/api/v1/task.go
```diff
@@ -100,108 +118,94 @@ func (h *TaskHandler) GetTask(c *gin.Context) {
...
 // @Failure 500 {object} ierr.ErrorResponse
-// @Router /tasks/{id}/process [post]
-func (h *TaskHandler) ProcessTask(c *gin.Context) {
-	id := c.Param("id")
-	if id == "" {
-		c.Error(ierr.WithError(nil).
-			WithHint("Please check the task id").
+// @Router /tasks/result [get]
+func (h *TaskHandler) GetTaskProcessingResult(c *gin.Context) {
+	workflowID := c.Query("workflow_id")
```
nkmishra1997, line 192: this API needs to work on task_id primarily 
and workflow_id can be optional

internal/service/file_provider.go
```diff
@@ -0,0 +1,259 @@
...
+	GetDownloadURL(ctx context.Context, fileURL string) (string, error)
+	GetProviderName() FileProviderType
+}
+
+type FileProviderType string
+
+const (
+	FileProviderTypeDirect      FileProviderType = "direct"
+	FileProviderTypeGoogleDrive FileProviderType = "google_drive"
+	FileProviderTypeS3          FileProviderType = "s3"
```
nkmishra1997, line 49: In future we might support using particular credentials (via s3 integration) for downloading private s3 files

internal/service/streaming_processor.go
```diff
@@ -0,0 +1,506 @@
...
+)
+
+// StreamingProcessor handles streaming processing of large files
+type StreamingProcessor struct {
+	Client           httpclient.Client
+	Logger           *logger.Logger
+	ProviderRegistry *FileProviderRegistry
+	CSVProcessor     *CSVProcessor
+	JSONProcessor    *JSONProcessor
+	WorkerPool       *workerpool.WorkerPool
```
nkmishra1997: I don't think we need to this as it is not getting used anywhere so can clean up in follow up PR

internal/service/streaming_processor.go
```diff
@@ -0,0 +1,506 @@
...
+		MaxRetries:     3,
+		RetryDelay:     5 * time.Second,
+		MaxErrors:      1000, // Stop processing after 1000 errors
+		BatchSize:      10,   // Update progress every 10 chunks
+	}
+}
+
+// detectFileType attempts to determine if the file is CSV or JSON
+func (sp *StreamingProcessor) detectFileType(content []byte) FileType {
+	// Skip BOM if present
```
nkmishra1997, line 91: remove unused method

internal/temporal/activities/task_activities.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Set context values using centralized utilities
+	ctx = types.SetTenantID(ctx, input.TenantID)
+	ctx = types.SetEnvironmentID(ctx, input.EnvironmentID)
+
+	// Start a goroutine to send heartbeats
+	heartbeatCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	go a.sendHeartbeats(heartbeatCtx, input.TaskID)
```
nkmishra1997: need to remove this

internal/temporal/activities/task_activities.go
nkmishra1997: bumping this up

internal/temporal/activities/task_activities.go
```diff
@@ -0,0 +1,118 @@
...
+	if err := input.Validate(); err != nil {
+		return nil, err
+	}
+
+	// Set context values using centralized utilities
+	ctx = types.SetTenantID(ctx, input.TenantID)
+	ctx = types.SetEnvironmentID(ctx, input.EnvironmentID)
+
+	// Start a goroutine to send heartbeats
+	heartbeatCtx, cancel := context.WithCancel(ctx)
```
nkmishra1997: remove

internal/service/streaming_processor.go
```diff
@@ -0,0 +1,506 @@
...
+				"max_retries": config.MaxRetries,
+			}).
+			Mark(ierr.ErrValidation)
+	}
+
+	return chunkResult, nil
+}
+
+// downloadFileStream downloads a file and returns a stream using retryable HTTP client
+func (sp *StreamingProcessor) downloadFileStream(ctx context.Context, t *task.Task) (io.ReadCloser, error) {
```
nkmishra1997, line 432: One problem with keeping everything in memory during the whole processing is that it is going to take RAM usage
Ideally we should store the file on the disk
then partially read the disk stored file in the stream processor and at the end of processing - delete the file

nkmishra1997 (approved):

nkmishra1997 (approved):

PR #586, opened 2025-09-27
fix(alert): reframe clean DB wallet transaction followed by alert logging

subratsahilgupta:


hiteshshimpi-55 (commented):
Looks good to me

PR #619, opened 2025-10-06
feat(wallets): calculate wallet balance based on allowed price type

omkar273:
## Summary by CodeRabbit

- New Features
  - Wallet balances now respect configured price types (All, Fixed, Usage), with usage-based charges included only when allowed.
  - Responses now include allowed price types for greater transparency.
- Improvements
  - Unpaid invoice totals are calculated based on the allowed price types, improving accuracy.
  - Backward compatibility: if no price types are configured, all price types are treated as allowed by default.
- Bug Fixes
  - Prevents unintended inclusion of usage charges when wallets are restricted to fixed-price items.

### DIFF
--- a/internal/service/invoice.go
+++ b/internal/service/invoice.go
@@ -1242,7 +1242,7 @@ func (s *invoiceService) GetUnpaidInvoicesToBePaid(ctx context.Context, customer
 	containsPriceTypeFixed := lo.Contains(allowedPriceTypes, types.WalletConfigPriceTypeFixed)
 	containsPriceTypeUsage := lo.Contains(allowedPriceTypes, types.WalletConfigPriceTypeUsage)
 
-	if containsPriceTypeAll && containsPriceTypeFixed || containsPriceTypeAll {
+	if containsPriceTypeUsage && containsPriceTypeFixed || containsPriceTypeAll {
 		// All allowed, include all
 		unpaidAmount = usageAmountTotal.Add(fixedAmountTotal)
 	} else if containsPriceTypeUsage {
--- a/internal/service/invoice.go
+++ b/internal/service/invoice.go
@@ -1238,16 +1238,17 @@ func (s *invoiceService) GetUnpaidInvoicesToBePaid(ctx context.Context, customer
 	var unpaidAmount decimal.Decimal
 
 	// Check if we need to filter by price type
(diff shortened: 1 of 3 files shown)


omkar273:
@coderabbitai review

internal/service/invoice.go
```diff
@@ -1186,15 +1186,16 @@ func (s *invoiceService) GetCustomerInvoiceSummary(ctx context.Context, customer
...
-	unpaidAmount := decimal.Zero
+	fixedAmountTotal := decimal.Zero
+	usageAmountTotal := decimal.Zero
 
 	filter := types.NewNoLimitInvoiceFilter()
 	filter.QueryFilter.Status = lo.ToPtr(types.StatusPublished)
 	filter.CustomerID = customerID
 	filter.InvoiceStatus = []types.InvoiceStatus{types.InvoiceStatusFinalized}
-	filter.SkipLineItems = true
+	filter.SkipLineItems = false
```
nkmishra1997, line 1198: can optimise this a bit by still skipping line items for `ALL` types of wallet config price type as in that case we will be unnecessarily fetching line items which can be expensive

internal/service/invoice.go
```diff
@@ -1216,10 +1217,44 @@ func (s *invoiceService) GetUnpaidInvoicesToBePaid(ctx context.Context, customer
...
+	} else if containsPriceTypeUsage {
+		// Usage-only wallet
+		unpaidAmount = usageAmountTotal
+	} else if containsPriceTypeFixed {
+		// Fixed-only wallet
+		unpaidAmount = fixedAmountTotal
+	} else {
+		// No price type filter, include all
+		unpaidAmount = usageAmountTotal.Add(fixedAmountTotal)
+	}
```
nkmishra1997, line 1257: This logic has to factor in and make sure it is never greater than the inv.AmountRemaining

Or maybe I am thinking - lets just get rid of unpaid invoices all together.
and only handle for current months usage. This is much more neater and easy to understand and less complex to implement.

In all cases I mean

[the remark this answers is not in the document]
nkmishra1997: @omkar273 either this kind of a logic or maybe lets just get rid of the pending invoice totals all together.

PR #1392, opened 2026-03-16
fix(subscription): add support addon associations cancellation and sub line item termination post cancellation

subratsahilgupta:
## Release Notes

* **Bug Fixes**
  * Addons are now properly cancelled and terminated when a subscription is cancelled, ensuring accurate tracking of addon status and associated line items.
  * Addon associations now reflect cancellation state with appropriate details and end dates.

* **Tests**
  * Added test coverage for addon cancellation scenarios.

### DIFF
--- a/internal/service/subscription_test.go
+++ b/internal/service/subscription_test.go
@@ -1471,6 +1471,96 @@ func (s *SubscriptionServiceSuite) TestCancelSubscription() {
 		})
 	})
 
+	s.Run("TestCancelSubscriptionWithAddons", func() {
+		ctx := s.GetContext()
+		subService := s.service.(*subscriptionService)
+
+		// Create subscription to cancel
+		subWithAddon := &subscription.Subscription{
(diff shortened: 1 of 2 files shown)


internal/service/subscription.go
```diff
@@ -4039,6 +4044,129 @@ func (s *subscriptionService) validateEntitlementCompatibility(ctx context.Conte
...
+		}
+		association := addonResp.AddonAssociation
+
+		// Skip if already has end date (already scheduled for removal)
+		if association.EndDate != nil && !association.EndDate.IsZero() {
+			logger.Debugw("addon association already has end date, skipping",
+				"addon_association_id", association.ID,
+				"end_date", association.EndDate)
+			continue
+		}
```
nkmishra1997, line 4096: @subratsahilgupta need to handle this - future addon expiry also needs to be cancelled

nkmishra1997: can retain the end date rather than skipping

PR #1415, opened 2026-03-20
feat(customer): add support for customer hierarchy

omkar273:
## PR Overview

---

### 🧩 Problem Statement

This PR solves three interconnected problems in the subscription system:

1. **No subscription hierarchy model** — the system had no first-class concept of parent/child subscription relationships. Customers sharing a plan under an enterprise account had to be managed ad-hoc with no guardrails.

2. **Analytics blind spot for child customers** — when usage analytics were requested for a child customer's external ID, usage was correctly found but **cost was always \$0**. Root cause: inherited subscriptions carry no line items (pricing lives on the parent subscription), so cost calculation had nothing to read.

3. **Stale API surface** — `CreateSubscriptionRequest` had six flat top-level inheritance fields mixed in with regular subscription fields. There was no way to add new children to an already-existing parent subscription after it was created. Additionally, two deprecated fields (`parent_customer_id` on Customer, `invoice_billing` on Subscription) were still being read/written in service code despite being superseded.

---

### 🛠 Dev Approach

The work was delivered in four sequential layers:

#### 1. Subscription Type Foundation
Introduced `subscription_type` enum (`standalone` | `parent` | `inherited`) on the Subscription entity. All subscriptions default to `standalone`. Added to Ent schema, domain model, filters, API responses, and ordering. Added `UsageCustomerIDs` / `CustomerIDsToInheritSubscription` to track which child customer IDs roll up into a parent subscription's usage.

#### 2. Subscription Inheritance Service
Built the core hierarchy engine:
- `createInheritedSubscriptions` — clones a parent subscription into N child subscriptions, each linked via `parent_subscription_id`
- `validateCustomerSubscriptionWorkflow` — prevents illegal states (e.g. a child customer already having a standalone subscription)
- `getInheritedSubscriptions` — fetches active/trialing/paused children for a parent
- Cascade logic: cancel/pause/resume on a parent propagates to all children

#### 3. Analytics Cost Fix
`fetchSubscriptions` in `feature_usage_tracking.go` was updated: when a subscription has type `INHERITED`, it now also fetches the parent subscription and appends it to the set. This gives the cost calculator access to the parent's line items (prices) so inherited customer analytics return correct costs instead of \$0.

#### 4. API Surface Cleanup
- `CreateSubscriptionRequest`: 6 flat inheritance fields replaced with a single `inheritance` object (`SubscriptionInheritanceConfig`)
- New endpoint: `POST /subscriptions/:id/inheritance/execute` — adds child customers to an existing parent subscription without recreating it
- Multi-customer analytics: `external_customer_ids[]` + `include_children=true` on usage endpoints
- Removed `parent_customer_id` from Customer (domain, DTO, service, repo, filter) — application-level only, DB column retained
- Removed `invoice_billing` from Subscription (type, DTO field, service fallback, validation blocks)

---

#### Subscription Hierarchy Creation
```
POST /subscriptions (with inheritance object)
         │
         ▼
  CreateSubscription
         │
         ├─ type == "parent"?
         │         │
         │         ▼
         │  resolveUsageCustomerIDs()
         │         │
         │         ├─ validate no STANDALONE conflicts
         │         │
         │         └─ createInheritedSubscriptions()
         │                   │
         │                   └─ for each child customer:
         │                        clone parent sub
         │                        set subscription_type = "inherited"
         │                        set parent_subscription_id = parentSub.ID
         │
         └─ type == "inherited"?  (ParentSubscriptionID set)
                   │
                   └─ validate parent exists and is type=parent
```

#### Execute Inheritance (Add Children to Existing Parent)
```
POST /subscriptions/:id/inheritance/execute
         │
         ▼
  Fetch parent subscription by :id
         │
         ├─ Not found → 404
         ├─ Not type=parent → 400
         │
         ▼
  Resolve customer IDs (internal or external)
         │
         ▼
  Fetch existing inherited children (active/trialing/paused)
         │
         ▼
  Filter out already-inherited customers
         │
         ├─ All already inherited → 200 (no-op)
         │
         ▼
  validateCustomerSubscriptionWorkflow()
         │
         ▼
  createInheritedSubscriptions()
         │
         ▼
  Return updated parent subscription
```

#### Analytics Cost Fix — Inherited Customer
```
GET /analytics?external_customer_id=child-ext-id
         │
         ▼
  fetchSubscriptions(childCustomerID)
         │
         ├─ finds INHERITED subscription
         │
         ▼
  detect subscription_type == INHERITED
         │
         ▼
  fetch parent subscription (via parent_subscription_id)
         │
         ▼
  append parent sub to subscription set
         │
         ▼
  cost calculator reads line items from parent sub ✅
  (previously: no line items on child → cost = $0 ❌)
```

---

### ⚠️ Breaking Changes

| Change | Impact |
|--------|--------|
| `CreateSubscriptionRequest` — 6 flat inheritance fields removed, replaced with `inheritance` object | Any caller passing `invoicing_customer_id`, `invoicing_customer_external_id`, `parent_subscription_id`, `customer_ids_to_inherit_subscription`, `external_customer_ids_to_inherit_subscription`, `invoice_billing` at the top level must move them under `inheritance: { ... }` |
| `invoice_billing` field removed entirely | Callers using `invoice_billing: "invoice_to_parent"` must switch to `inheritance.invoicing_customer_id` or `inheritance.invoicing_customer_external_id` — no auto-inference fallback |
| `parent_customer_id` / `parent_customer_external_id` removed from Create/Update Customer request | Fields are silently ignored if sent; no error, but the value is no longer stored or returned |
| `parent_customer` removed from `GET /customers/:id` response | Callers reading `parent_customer` in the response will receive `null` |
| `execute-subscription-inheritance` requires exactly one of `customer_ids` or `external_customer_ids` | Sending both → 400 validation error |

> Note: The DB column `parent_customer_id` is retained — only the application layer was cleaned up.

---

#### Subscription Inheritance — Creation
- ✅ Create parent subscription with `customer_ids_to_inherit_subscription` — child subs created with `type=inherited`, linked to parent
- ✅ Create parent subscription with `external_customer_ids_to_inherit_subscription` — external IDs resolved to internal before creating children
- ✅ Providing both `customer_ids` and `external_customer_ids` → 400 validation error
- ✅ Child customer with existing STANDALONE subscription → conflict error
- ✅ Child customer with existing INHERITED subscription (different parent) → conflict error
- ✅ Cascade cancel: cancelling parent → all children cancelled
- ✅ Cascade pause: pausing parent → all children paused
- ✅ Cascade resume: resuming parent → all children resumed

#### Execute Inheritance Endpoint
- ✅ Add new children to existing parent subscription
- ✅ Already-inherited customer in request → silently skipped (no duplicate created)
- ✅ Previously-cancelled child re-added → new inherited sub created (intentional)
- ✅ Parent not found → 404
- ✅ Subscription is not type=parent → 400
- ✅ Both arrays provided → 400
- ✅ Both arrays empty → 400
- ✅ All customers already inherited → 200 no-op, returns parent unchanged

#### Analytics — Multi-Customer & Inherited
- ✅ `include_children=true` aggregates parent + all child usage into one merged total
- ✅ `external_customer_ids[]=a&external_customer_ids[]=b` returns aggregated usage for multiple customers
- ✅ Child customer analytics: usage correctly found AND cost is non-zero (parent line items used)
- ✅ Standalone customer analytics: unaffected by hierarchy changes
- ✅ Parent customer analytics with children: cost correctly aggregated across hierarchy

#### Deprecated Field Removal
- ✅ `CreateCustomerRequest` with `parent_customer_id` — field silently ignored
- ✅ `UpdateCustomerRequest` with `parent_customer_id` — field silently ignored
- ✅ `GET /customers/:id` — no `parent_customer` key in response
- ✅ `CreateSubscriptionRequest` with `invoice_billing` in `inheritance` — field silently ignored
- ✅ `InvoiceBilling` type fully absent — zero references in application code

#### Validation
- ✅ `invoicing_customer_id` and `invoicing_customer_external_id` mutually exclusive → 400
- ✅ `parent_subscription_id` references non-existent subscription → 404
- ✅ `parent_subscription_id` references non-parent-type subscription → 400
- ✅ Invoicing customer not active → 400

----

## Summary by CodeRabbit

* **New Features**
  * Subscription hierarchy: new subscription types (standalone, parent, inherited) with defaults; subscription_type included in APIs, responses, filters, and ordering.
  * Inheritance management: POST /subscriptions/{id}/inheritance/execute to add inherited children; creation/update APIs and lifecycle actions cascade/validate across hierarchy.
  * Multi-customer analytics: usage endpoints accept multiple external_customer_ids and include_children with deduped aggregation.

* **Bug Fixes**
  * Fixed analytics/cost calculations for inherited subscriptions by including parent line items in analytics.

* **Docs & Tests**
  * Updated OpenAPI/docs and added extensive tests for hierarchy, inheritance, analytics, and entitlement scenarios.


internal/repository/clickhouse/aggregators.go
```diff
@@ -72,6 +73,38 @@ func getDeduplicationKey() string {
...
+	if len(unique) == 1 {
+		return fmt.Sprintf("AND external_customer_id = '%s'", unique[0])
+	}
+	
+	// Multiple customer IDs - use IN clause
+	quotedIDs := make([]string, len(unique))
+	for i, id := range unique {
+		quotedIDs[i] = fmt.Sprintf("'%s'", id)
+	}
+	return fmt.Sprintf("AND external_customer_id IN (%s)", strings.Join(quotedIDs, ", "))
```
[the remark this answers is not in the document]
omkar273: We dont have external ids with "" or \ so it wont be a issue

internal/repository/clickhouse/aggregators.go
```diff
@@ -343,7 +370,7 @@ func (a *CountAggregator) GetQuery(ctx context.Context, params *events.UsagePara
 
 	externalCustomerFilter := ""
 	if params.ExternalCustomerID != "" {
-		externalCustomerFilter = fmt.Sprintf("AND external_customer_id = '%s'", params.ExternalCustomerID)
+		externalCustomerFilter = buildExternalCustomerFilter(params)
 	}
```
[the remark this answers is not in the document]
omkar273: Handled

internal/service/billing.go
```diff
@@ -2920,6 +2920,24 @@ func (s *billingService) GetCustomerUsageSummary(ctx context.Context, customerID
...
+		}
+
+		customerIDs, err := service.GetAggregatedCustomerIDs(ctx, sub)
+
+		if err != nil {
+			return nil, err
+		}
+
+		subscriptionCustomerIDsMap[subscriptionID] = customerIDs
+	}
```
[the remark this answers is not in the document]
omkar273: Handled

omkar273 (commented):
@coderabbitai review

internal/api/dto/subscription.go
```diff
@@ -366,6 +366,16 @@ type CreateSubscriptionRequest struct {
 
 	// Enable Commitment True Up Fee
 	EnableTrueUp bool `json:"enable_true_up"`
+
+	// UsageCustomerIDs specifies which child customers' usage should be aggregated
+	// on this subscription. When set the subscription becomes a PARENT type and
+	// skeleton INHERITED subscriptions are created for each listed customer.
+	// Mutually exclusive with InvoiceBilling/InvoicingCustomerID.
+	// nil slice = JSON omitted (auto-detect children); non-nil = explicit list (empty slice allowed).
+	UsageCustomerIDs []string `json:"usage_customer_ids,omitempty"`
```
nkmishra1997: CustomerIDsToInheriSubscription []String

internal/api/dto/subscription.go
```diff
@@ -394,6 +404,31 @@ type UpdateSubscriptionRequest struct {
 
 	// ParentSubscriptionID sets or clears the parent subscription. Omit to leave unchanged; send "" to clear.
 	ParentSubscriptionID *string `json:"parent_subscription_id,omitempty"`
+
+	// UsageCustomerIDs updates the set of child customers whose usage is aggregated on this
+	// subscription. Omit the field (nil) to leave the current set unchanged. Provide a non-nil
+	// slice to declare the full desired set — all currently tracked customers must be present or
+	// an error is returned (removals require cancelling the subscription). Include additional IDs
+	// to add new children; duplicate IDs are silently deduplicated.
+	UsageCustomerIDs *[]string `json:"usage_customer_ids,omitempty"`
```
nkmishra1997: same rename here

internal/api/dto/subscription.go
```diff
@@ -1535,6 +1579,10 @@ type GetUsageBySubscriptionRequest struct {
 	// Source indicates the caller context. When "invoice_creation", ClickHouse queries use FINAL.
 	// Optional; omit for default (no FINAL).
 	Source string `json:"-"`
+	// OverrideCustomerIDs, when non-empty, replaces the auto-detected customer ID list used for
+	// usage aggregation. Intended for internal use only (e.g. per-child breakdown within a PARENT
+	// subscription). Not exposed via the HTTP API.
+	OverrideCustomerIDs []string `json:"-"`
```
nkmishra1997: @nkmishra1997 revisit

internal/domain/events/repository.go
```diff
@@ -82,8 +82,9 @@ type UsageAnalytic struct {
 }
 
 type UsageParams struct {
-	ExternalCustomerID string                `json:"external_customer_id"`
-	CustomerID         string                `json:"customer_id"`
+	ExternalCustomerID  string                `json:"external_customer_id"`
+	ExternalCustomerIDs []string              `json:"external_customer_ids"`
+	CustomerID          string                `json:"customer_id"`
```
nkmishra1997, line 90: formatting

internal/repository/clickhouse/feature_usage.go
```diff
@@ -2031,33 +2057,33 @@ func (r *FeatureUsageRepository) GetFeatureUsageBySubscription(ctx context.Conte
...
-			AND customer_id = ?
+			AND %s
 			AND environment_id = ?
 			AND tenant_id = ?
 			AND "timestamp" >= ?
 			AND "timestamp" < ?
 			AND sign != 0
 		GROUP BY sub_line_item_id, feature_id, meter_id, price_id
-	`, strings.Join(aggColumns, ",\n\t\t\t"), tableRef)
+	`, strings.Join(aggColumns, ",\n\t\t\t"), tableRef, customerFilter)
```
nkmishra1997, line 2059: NIT: could have been simplified to make it easy to read

internal/repository/clickhouse/feature_usage.go
nkmishra1997, line 2111: bascially because of these aggregations we need n separate queries whenever we want the breakdown for n customers

internal/repository/clickhouse/feature_usage.go
```diff
@@ -1995,15 +1995,15 @@ func (r *FeatureUsageRepository) getAnalyticsPoints(
 
 // GetFeatureUsageBySubscription gets usage data for a subscription using a single optimized query.
 // When opts.Source is InvoiceCreation, the query uses FINAL for correct ReplacingMergeTree deduplication.
-func (r *FeatureUsageRepository) GetFeatureUsageBySubscription(ctx context.Context, subscriptionID, customerID string, startTime, endTime time.Time, aggTypes []types.AggregationType, opts *events.GetFeatureUsageBySubscriptionOpts) (map[string]*events.UsageByFeatureResult, error) {
+func (r *FeatureUsageRepository) GetFeatureUsageBySubscription(ctx context.Context, subscriptionID string, customerIDs []string, startTime, endTime time.Time, aggTypes []types.AggregationType, opts *events.GetFeatureUsageBySubscriptionOpts) (map[string]*events.UsageByFeatureResult, error) {
```
nkmishra1997: Some combinations that internal functions will use while calling this dumb but expensive function:
S1 -> parent / S2 -> inherited

1. S1 + [C1]
2. S1 + [C1,C2,C3] (asked to include child usage without breakdown - WE NEVER GIVE WITH BREAKDOWN)
3. S2 + [C2] (not possible) -> empty result if called
4. S2 + [C2, C3] (not possible) -> empty result if called
5. S1 + [C2] (possible) basically child customer's analytics tab)
6. S1 + [C2,C3] (not possible) -> will give all child usage in a way

nkmishra1997: One critical case
C2 has one subscription S3 (standalone) -> not allowed

So default behaviour for C2 analytics -> show their contribution in the parent's subscriptio
AND FOR NOW WE DO NOT SUPPORT querying parent usage directly -> call the parent's usage 
This might get supported in future if customers request but then internally it is basically S1 + [C1,C2,C3]  AND not without a breakdown separately


REMOVE BREAKDOWN COMPLETELY

nkmishra1997: Simplify and remove breakdown even from entitlements etc

internal/service/subscription.go
```diff
@@ -1593,12 +1623,38 @@ func (s *subscriptionService) UpdateSubscription(ctx context.Context, subscripti
 
 	subscription.CancelAtPeriodEnd = req.CancelAtPeriodEnd
 
-	// Update the subscription in the database
-	err = s.SubRepo.Update(ctx, subscription)
+	// --- usage_customer_ids: add new child customers to an existing subscription ---
```
nkmishra1997: add validation to allow parent_subscription_id update only in case of standalone subscriptions

internal/api/dto/subscription.go
nkmishra1997: since we are not giving breakdown -> this just changes to customer_ids array

internal/service/subscription.go
```diff
@@ -1593,12 +1623,38 @@ func (s *subscriptionService) UpdateSubscription(ctx context.Context, subscripti
...
+	// Omitting the field (nil) leaves the current set unchanged.
+	// Providing a non-nil slice declares the full desired set; any existing customer
+	// not present in the new list is rejected — removals require cancellation.
+	var addedCustomerIDs []string
```
nkmishra1997: rename the parameter to add_external_customer_ids_to_inherit

internal/service/subscription.go
```diff
@@ -1609,6 +1665,145 @@ func (s *subscriptionService) UpdateSubscription(ctx context.Context, subscripti
...
+//   - Duplicate IDs in desiredIDs are silently deduplicated.
+//   - Returns the net-new customer IDs (desiredIDs minus already-inherited).
+func (s *subscriptionService) resolveUsageCustomerIDsUpdate(
+	ctx context.Context,
+	sub *subscription.Subscription,
+	desiredIDs []string,
+) ([]string, error) {
+	// 1. Subscription must be live.
+	switch sub.SubscriptionStatus {
+	case types.SubscriptionStatusActive, types.SubscriptionStatusTrialing:
```
nkmishra1997: allow for draft as well

internal/service/subscription.go
```diff
@@ -112,6 +112,23 @@ func (s *subscriptionService) CreateSubscription(ctx context.Context, req dto.Cr
 		}
 	}
 
+	// --- Customer hierarchy: resolve usage_customer_ids and set SubscriptionType ---
+	usageCustomerIDs, err := s.resolveUsageCustomerIDs(ctx, customer, &req)
```
nkmishra1997: remove and create new API:
1. Check if standalone -> convert to parent
2. Also do not allow if standalone and has invoice_customer_id set -> ask to remove that first (for which there is no way right now but we can add later if this use case comes)
3. only allow adding new customers right now (external customer ids only)
4. And will validate the customer getting added is valid and does not have any other subscription (standalone/parent/inherited doesn't matter)

internal/service/subscription.go
```diff
@@ -112,6 +112,23 @@ func (s *subscriptionService) CreateSubscription(ctx context.Context, req dto.Cr
...
+	if err != nil {
+		return nil, err
```
nkmishra1997, line 130: Create subscription feedback:
1. Do not automatically pick all child customer ids etc
2. Just remove option from UI to mark parent/child (We should discontinue that flow immediately)
3. give option on UI to add invoicing_customer_id and similarly add customer to inherit ids (invoicing config) with proper checks to make the invoicing_customer_id  and inheriting logic mutually exclusive
4. And eventually remove the parent_customer_id from customer create/update apis (mention as breaking change in the release) 2.0.0 - 2.1.0 (breaking change)
5. And eventually remove the invoice_billing from create subscription (do the breaking changes in same release)

internal/service/subscription.go
```diff
@@ -4981,7 +5208,27 @@ func (s *subscriptionService) GetFeatureUsageBySubscription(ctx context.Context,
 	opts := &events.GetFeatureUsageBySubscriptionOpts{
 		Source: types.UsageSource(req.Source),
 	}
-	usageResults, err := s.FeatureUsageRepo.GetFeatureUsageBySubscription(ctx, req.SubscriptionID, customer.ID, usageStartTime, usageEndTime, aggTypes, opts)
+	// Determine customer IDs for usage query.
+	// If the caller explicitly provides OverrideCustomerIDs (e.g. per-child breakdown), use those
+	// directly. Otherwise auto-detect: for PARENT subs include all inherited children; for all
+	// other types scope to the subscription's own customer.
+	var usageCustomerIDs []string
+	if len(req.OverrideCustomerIDs) > 0 {
```
nkmishra1997: this should not be required inside service
service to resolve standalone vs inherited vs parent customer cases

internal/service/subscription.go
```diff
@@ -2089,29 +2291,34 @@ func (s *subscriptionService) GetUsageBySubscription(ctx context.Context, req *d
...
-	// Create a map for fast event name lookup
-	eventNameExists := make(map[string]bool, len(distinctEventNames))
-	for _, eventName := range distinctEventNames {
-		eventNameExists[eventName] = true
+	// Performance optimization: Get distinct event names across all relevant customers
+	// to filter out meters that have no events, reducing processing from potentially
+	// 400-500 meters down to only 5-7 that have actual usage
+	distinctEventNameSet := make(map[string]bool)
+	for _, extID := range aggregatedExternalIDs {
+		names, derr := s.EventRepo.GetDistinctEventNames(ctx, extID, usageStartTime, usageEndTime)
```
nkmishra1997: this should also take external_customer_ids

internal/service/subscription.go
```diff
@@ -2089,29 +2291,34 @@ func (s *subscriptionService) GetUsageBySubscription(ctx context.Context, req *d
 		"end_time", usageEndTime,
 		"metered_line_items", len(priceIDs))
 
-	// Performance optimization: Get distinct event names for this customer
-	// to filter out meters that have no events, reducing processing from potentially
-	// 400-500 meters down to only 5-7 that have actual usage
-	distinctEventNames, err := s.EventRepo.GetDistinctEventNames(ctx, customer.ExternalID, usageStartTime, usageEndTime)
+	// For PARENT-type subscriptions, aggregate usage from all child customers as well
+	inheritanceSvc := NewSubscriptionInheritanceService(s.ServiceParams)
+	aggregatedExternalIDs, err := inheritanceSvc.GetAggregatedCustomerIDs(ctx, subscription)
```
nkmishra1997: clean up

just make a helper method inside subscription service
GetCustomerIDs(subcsription)
GetExternalCustomerIDs(subcsription)
basis subscription type -> it will resolve customer ids

Be mindful about every extra DB query here

internal/service/subscription.go
nkmishra1997: should be just one query

internal/api/dto/events.go
```diff
@@ -95,16 +96,17 @@ type GetUsageRequest struct {
...
-	CustomerID         string              `form:"customer_id" json:"customer_id" example:"customer456"`
-	StartTime          time.Time           `form:"start_time" json:"start_time" example:"2024-11-09T00:00:00Z"`
-	EndTime            time.Time           `form:"end_time" json:"end_time" example:"2024-12-09T00:00:00Z"`
-	WindowSize         types.WindowSize    `form:"window_size" json:"window_size"`
-	BucketSize         types.WindowSize    `form:"bucket_size" json:"bucket_size,omitempty" example:"HOUR"` // Optional, only used for MAX aggregation with windowing
-	Filters            map[string][]string `form:"filters,omitempty" json:"filters,omitempty"`
+	MeterID             string              `form:"meter_id" json:"meter_id" binding:"required" example:"123"`
+	PriceID             string              `form:"-" json:"-"` // this is just for internal use to store the price id
+	Meter               *meter.Meter        `form:"-" json:"-"` // caller can set this in case already fetched from db to avoid extra db call
+	ExternalCustomerID  string              `form:"external_customer_id" json:"external_customer_id" example:"user_5"`
```
nkmishra1997, line 134: double check to make sure atleast internal code apart from the router handler does not use this parameter
all internal methods even if just one external id - should use the array

internal/service/subscription.go
```diff
@@ -2448,6 +2650,10 @@ func (s *subscriptionService) UpdateBillingPeriods(ctx context.Context) (*dto.Su
 				Status: lo.ToPtr(types.StatusPublished),
 			},
 			SubscriptionStatus: []types.SubscriptionStatus{types.SubscriptionStatusActive},
+			SubscriptionTypes: []types.SubscriptionType{
+				types.SubscriptionTypeStandalone,
```
nkmishra1997, line 2515: check in temporal cron also to use the same concept

internal/domain/events/repository.go
omkar273: Done

internal/api/dto/events.go
omkar273: Done

internal/api/dto/subscription.go
omkar273: Fixed — renamed to CustomerIDsToInheritSubscription throughout the DTO, service, and all test files.

internal/api/dto/subscription.go
omkar273: Fixed — renamed to CustomerIDsToInheritSubscription and external_customer_ids_to_inherit_subscription everywhere.

internal/api/dto/subscription.go
omkar273: Revisited and addressed — the flat inheritance fields have been fully replaced with a nested inheritance object (SubscriptionInheritanceConfig). The old flat fields are gone.

internal/repository/clickhouse/feature_usage.go
omkar273: Acknowledged — simplified the customerIDs resolution path. The logic is now consolidated in resolveUsageCustomerIDs which handles both internal and external ID resolution in a single pass.

internal/repository/clickhouse/feature_usage.go
omkar273: Acknowledged — this is a known architectural tradeoff for this iteration. The n-query pattern is needed for per-customer breakdown. Since we never return breakdown (merged total only per your feedback), the upstream call always passes a single unified ID set, keeping it to 1 query in practice.

internal/repository/clickhouse/feature_usage.go
omkar273: Fully aligned with these constraints — the implementation enforces exactly this: S2+[C2] (child-only) always routes through the parent sub for cost calculation (analytics cost fix). The direct query-parent use case is not surfaced in the API and is handled internally only.

internal/service/subscription.go
omkar273: Added validation — parent_subscription_id can only be set on STANDALONE subscriptions (or when creating a new one). Updating it on an already-PARENT or INHERITED subscription is rejected with a 400.

internal/api/dto/subscription.go
omkar273: Updated — usage customer IDs are now a flat array (customer_ids_to_inherit_subscription / external_customer_ids_to_inherit_subscription) inside the inheritance object. No breakdown field exposed.

internal/service/subscription.go
omkar273: Used external_customer_ids_to_inherit_subscription in the final implementation — kept the full descriptive name rather than abbreviating to make intent explicit in the API contract.

internal/service/subscription.go
omkar273: Draft subscriptions are now included — getInheritedSubscriptions fetches active, trialing, paused, and draft children. The execute endpoint correctly skips already-inherited customers regardless of their draft/active state.

internal/service/subscription.go
omkar273: Done — implemented as POST /subscriptions/:id/inheritance/execute. It: (1) verifies the target subscription is already PARENT type, (2) accepts only new child customer IDs to add (no removal in this iteration), (3) skips customers already in the hierarchy, (4) validates no standalone conflicts. The old standalone->parent mutation is not exposed.

internal/service/subscription.go
omkar273: All addressed: (1) removed automatic child detection — callers must explicitly pass customer IDs, (2) removed parent/child flag from the create flow — it's now purely driven by the inheritance object in the request, (3) invoicing_customer_id and external customer IDs are explicit fields in the inheritance object, (4) the execute endpoint handles post-creation additions. Removal API is out of scope for this PR.

internal/service/subscription.go
omkar273: Cleaned up — moved all hierarchy resolution into resolveUsageCustomerIDs in the service. The handler layer passes raw customer IDs; the service resolves standalone/inherited/parent semantics internally.

internal/service/subscription.go
omkar273: Added — both internal (customer_ids_to_inherit_subscription) and external (external_customer_ids_to_inherit_subscription) ID arrays are supported on both the create and execute-inheritance endpoints. Exactly one of the two must be provided.

internal/service/subscription.go
omkar273: Refactored — hierarchy resolution is now a clean helper resolveUsageCustomerIDs(ctx, req) that returns the resolved []string of customer IDs based on subscription type. Subscription service consumers get the IDs without knowing the resolution details.

internal/service/subscription.go
omkar273: Consolidated to a single query — validateCustomerSubscriptionWorkflow now fetches all existing subscriptions for all candidate customer IDs in one batch query using CustomerIDsIn filter, rather than looping per customer.

internal/api/dto/events.go
omkar273: Verified — all internal service code uses the ExternalCustomerIDs []string array form. The singular ExternalCustomerID is only mapped at the router/handler layer for backward-compat with existing API consumers and immediately merged into the array before entering the service layer.

internal/service/subscription.go
omkar273: Noted — the temporal billing cron (UpdateBillingPeriods activity) already skips inherited subscriptions in processSubscriptionPeriod so they don't double-bill. Full temporal hierarchy propagation is tracked as follow-up work.

docs/superpowers/plans/2026-03-24-inherited-subscription-analytics-cost-fix.md
```diff
@@ -0,0 +1,350 @@
...
+```
+
+Replace the existing import block at the top of `feature_usage_tracking_analytics_test.go` with the above.
+
+- [ ] **Step 2: Run the tests to confirm they fail**
+
+```bash
+cd /Users/omkar/Developer/source-code/flexprice/flexprice
+go test ./internal/service/ -run TestFetchParentSubscriptions -v 2>&1 | head -30
+```
```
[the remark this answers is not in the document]
omkar273: Acknowledged — the absolute paths in the plan docs are artefacts of the AI-assisted planning tooling that generated them. They are internal dev docs (docs/superpowers/) and not meant for contributor execution. Will use repo-relative paths in future plan docs.

docs/superpowers/plans/2026-03-24-usage-analytics-multi-customer.md
```diff
@@ -0,0 +1,629 @@
...
+  	// IncludeChildren when true folds child customers' usage into the single aggregated total.
+  	// Default: false.
+  	IncludeChildren bool `json:"include_children,omitempty"`
+  }
+  ```
+
+- [ ] **Verify it compiles.**
+  ```bash
+  cd /Users/omkar/Developer/source-code/flexprice/flexprice && go build ./internal/api/dto/...
+  ```
```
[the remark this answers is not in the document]
omkar273: Same as above — acknowledged, internal planning docs only. Future plans will use repo-relative commands.

docs/superpowers/specs/2026-03-24-inherited-subscription-analytics-cost-fix-design.md
```diff
@@ -0,0 +1,145 @@
...
+analytics = fetchAnalytics(ctx, params)
+// → ClickHouse returns events with sub_line_item_id = "parent-line-item-id"
+
+calculateCosts(data)
+// → looks up "parent-line-item-id" in SubscriptionLineItems → NOT FOUND → cost = $0
+
+fetchSubscriptionPrices(data)
+// → iterates data.Subscriptions to collect PriceIDs from line items → none found → Prices map empty
+// → calculateCosts looks up Prices["parent-price-id"] → NOT FOUND → cost = $0 (secondary cause)
+```
```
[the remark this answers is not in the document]
omkar273: Acknowledged — minor doc formatting issue in internal spec files. Will add language identifiers to fenced blocks in future specs.

docs/superpowers/specs/2026-03-24-usage-analytics-multi-customer-design.md
```diff
@@ -0,0 +1,251 @@
...
+One-line normalisation at the very top of V1's `GetDetailedUsageAnalytics`, before any existing logic:
+
+```go
+if req.ExternalCustomerID == "" && len(req.ExternalCustomerIDs) > 0 {
+    req.ExternalCustomerID = req.ExternalCustomerIDs[0]
+}
+```
+
+V1 is single-customer only. Sending multiple IDs to V1 silently uses the first one — acceptable since V1 is a legacy path being phased out and does not support hierarchy or multi-customer aggregation.
+
```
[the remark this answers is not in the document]
omkar273: Acknowledged — the V1 path logs a warning when multiple IDs are provided and falls back to the first non-empty one. The spec wording has been noted for correction in a docs-only follow-up.

docs/swagger/docs.go
```diff
@@ -15448,6 +15467,10 @@ const docTemplate = `{
                         "type": "string"
                     }
                 },
+                "include_children": {
+                    "description": "IncludeChildren when true folds child customers' usage into the single aggregated total.\nDefault: false.",
+                    "type": "boolean"
```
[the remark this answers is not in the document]
omkar273: Acknowledged — will add default: false to the include_children schema property in a swagger-only follow-up commit.

ent/schema/subscription.go
```diff
@@ -195,6 +195,13 @@ func (Subscription) Fields() []ent.Field {
 			Nillable().
 			GoType(types.PaymentTerms("")).
 			Comment("Payment terms for invoice due date (e.g. 15 NET, 30 NET, 45 NET, 60 NET, 75 NET, 90 NET)"),
+		field.String("subscription_type").
+			SchemaType(map[string]string{
+				"postgres": "varchar(20)",
+			}).
+			Default(string(types.SubscriptionTypeStandalone)).
+			GoType(types.SubscriptionType("")).
+			Comment("Subscription type within a customer hierarchy (standalone, parent, inherited)"),
```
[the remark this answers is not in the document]
omkar273: Valid point — indexes for subscription_type and parent_subscription_id scoped to tenant+environment are needed for production scale. Tracked as a follow-up migration. Not blocking this PR as the feature is not yet in production load.

internal/api/dto/subscription.go
```diff
@@ -402,6 +417,16 @@ type UpdateSubscriptionRequest struct {
...
 
+// Validate checks UpdateSubscriptionRequest fields for basic structural validity.
+func (r *UpdateSubscriptionRequest) Validate() error {
+
+	if err := validator.ValidateRequest(r); err != nil {
+		return err
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
omkar273: Acknowledged — explicit status enum validation in UpdateSubscriptionRequest.Validate() is missing. Tracked as a follow-up hardening item. The service layer does validate subscription state before applying updates.

internal/repository/ent/customer.go
```diff
@@ -242,6 +242,94 @@ func (r *customerRepository) List(ctx context.Context, filter *types.CustomerFil
...
+			ON child_sub.parent_subscription_id = parent_sub.id
+			AND child_sub.tenant_id = parent_sub.tenant_id
+			AND child_sub.environment_id = parent_sub.environment_id
+		WHERE parent_sub.customer_id = $1
+			AND parent_sub.subscription_type = $2
+			AND child_sub.subscription_type = $3
+			AND parent_sub.tenant_id = $4
+			AND parent_sub.environment_id = $5
+			AND child_sub.customer_id <> $1
+	`
```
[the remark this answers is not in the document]
omkar273: Valid — the ListChildrenFromInheritedSubscriptions query should filter out cancelled/archived inherited subscriptions. The current implementation does filter by status=published on the customer side but not on the subscription side. Tracked as a follow-up fix.

internal/service/customer_test.go
```diff
@@ -350,6 +351,75 @@ func (s *CustomerServiceSuite) TestUpdateCustomer() {
...
+				Status: types.StatusPublished,
+			},
+		}
+	}
+
+	s.Run("standalone_active_subscription_allows_parent_change", func() {
+		s.SetupTest()
+		s.NoError(s.GetStores().CustomerRepo.Create(s.ctx, baseCustomer("parent-h")))
+		s.NoError(s.GetStores().CustomerRepo.Create(s.ctx, baseCustomer("child-h")))
+		s.NoError(s.GetStores().SubscriptionRepo.Create(s.ctx, baseSub("sub-sa", "child-h", "", types.SubscriptionStatusActive)))
```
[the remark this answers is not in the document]
omkar273: Fixed in this PR — the TestUpdateCustomer_ParentCustomerID test was deleted entirely as part of the ParentCustomerID removal. The standalone subscription type tests use types.SubscriptionTypeStandalone explicitly.

internal/service/event_post_processing.go
```diff
@@ -953,6 +953,17 @@ func (s *eventPostProcessingService) GetUsageAnalytics(ctx context.Context, tena
...
+	// V1 is single-customer only. If ExternalCustomerID is absent but ExternalCustomerIDs
+	// is provided, use the first entry so V1 callers aren't broken.
+	// Warn when multiple IDs are passed so operators can detect misuse.
+	if req.ExternalCustomerID == "" && len(req.ExternalCustomerIDs) > 0 {
+		if len(req.ExternalCustomerIDs) > 1 {
+			s.Logger.WarnwCtx(ctx, "V1 analytics does not support multiple customer IDs; using first entry only",
+				"external_customer_ids", req.ExternalCustomerIDs)
+		}
+		req.ExternalCustomerID = req.ExternalCustomerIDs[0]
+	}
```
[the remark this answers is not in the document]
omkar273: Acknowledged — the V1 fallback should skip blank entries and only fail when all IDs are empty. Tracked as a minor hardening item for follow-up.

internal/service/subscription.go
```diff
@@ -2478,6 +2522,10 @@ func (s *subscriptionService) UpdateBillingPeriods(ctx context.Context) (*dto.Su
...
+				types.SubscriptionTypeParent,
+			},
```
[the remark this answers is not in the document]
omkar273: Acknowledged — inherited subscriptions are intentionally excluded from UpdateBillingPeriods in this iteration. Billing for the hierarchy flows through the parent subscription only. The inherited rows are usage-tracking stubs; they do not generate independent invoices or billing periods. Full period sync for inherited rows is tracked as follow-up.

internal/testutil/inmemory_subscription_store.go
```diff
@@ -66,6 +66,42 @@ func subscriptionFilterFn(ctx context.Context, sub *subscription.Subscription, f
...
+
+	// Filter by parent subscription IDs (inherited subscription lookups)
+	if len(f.ParentSubscriptionIDs) > 0 {
+		if sub.ParentSubscriptionID == nil {
+			return false
+		}
+		if !lo.Contains(f.ParentSubscriptionIDs, *sub.ParentSubscriptionID) {
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
omkar273: Acknowledged — the in-memory store's subscriptionFilterFn doesn't narrow on SubscriptionIDs or InvoicingCustomerIDs. Tests that use those selectors may over-return rows. Tracked as an in-memory store hardening follow-up.

internal/types/subscription.go
```diff
@@ -40,6 +40,49 @@ func (i InvoiceBilling) Validate() error {
...
+}
+
+func (t SubscriptionType) String() string {
+	return string(t)
+}
+
+func (t SubscriptionType) Validate() error {
+	if t == "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
omkar273: Acknowledged — the filter validator should reject empty-string entries in subscription_types[]. Tracked as a minor validation follow-up.

docs/swagger/swagger-3-0.json
```diff
@@ -24308,6 +24357,13 @@
             "type": "string",
             "description": "CustomerID filters by customer ID"
           },
+          "customer_ids": {
+            "type": "array",
+            "description": "CustomerIDs filters by multiple customer IDs (used for hierarchy lookups)",
+            "items": {
+              "type": "string"
+            }
+          },
```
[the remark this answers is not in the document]
omkar273: Acknowledged — the filter exposes both singular CustomerID and plural CustomerIDs. They union — both are applied as IN predicates. Will clarify this in the schema description.

docs/swagger/swagger.yaml
```diff
@@ -3164,6 +3204,13 @@ definitions:
       external_customer_id:
         example: customer456
         type: string
+      external_customer_ids:
+        example:
+        - customer456
+        - customer789
+        items:
+          type: string
+        type: array
```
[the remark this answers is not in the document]
omkar273: Acknowledged — ExternalCustomerID (singular) is merged into ExternalCustomerIDs at the router layer before entering the service. The merged array is the authoritative input. Will document this in the swagger description.

docs/swagger/swagger.yaml
```diff
@@ -5180,6 +5227,11 @@ definitions:
         $ref: '#/definitions/types.Status'
       subscription_status:
         $ref: '#/definitions/types.SubscriptionStatus'
+      subscription_type:
+        allOf:
+        - $ref: '#/definitions/types.SubscriptionType'
+        description: SubscriptionType categorises this subscription within a customer
+          hierarchy.
```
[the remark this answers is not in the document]
omkar273: Good catch — subscription_type describes the subscription-level hierarchy, not a customer relationship. Will correct the wording in the swagger description.

internal/repository/ent/subscription.go
```diff
@@ -92,6 +92,7 @@ func (r *subscriptionRepository) Create(ctx context.Context, sub *domainSub.Subs
 		SetNillableInvoicingCustomerID(sub.InvoicingCustomerID).
 		SetNillableParentSubscriptionID(sub.ParentSubscriptionID).
 		SetNillablePaymentTerms(sub.PaymentTerms).
+		SetSubscriptionType(sub.SubscriptionType).
```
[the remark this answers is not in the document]
omkar273: Acknowledged — the Ent runtime.go sets the subscription_type default on new rows. Existing rows without the field will return an empty string, which the service treats as SubscriptionTypeStandalone. Explicit migration to backfill existing rows is tracked as follow-up.

internal/repository/ent/subscription.go
```diff
@@ -570,11 +572,21 @@ func (o *SubscriptionQueryOptions) applyEntityQueryOptions(_ context.Context, f
 		query = query.Where(subscription.ParentSubscriptionIDIn(f.ParentSubscriptionIDs...))
 	}
 
+	// Apply subscription type filter
+	if len(f.SubscriptionTypes) > 0 {
+		query = query.Where(subscription.SubscriptionTypeIn(f.SubscriptionTypes...))
+	}
```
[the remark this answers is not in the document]
omkar273: Acknowledged — legacy rows with empty subscription_type will be missed by predicates matching SubscriptionTypeStandalone explicitly. The service layer already normalises empty to standalone on read. A backfill migration to set subscription_type='standalone' on all existing rows will fix this permanently and is tracked as follow-up.

internal/service/customer_test.go
```diff
@@ -350,6 +351,75 @@ func (s *CustomerServiceSuite) TestUpdateCustomer() {
...
+		parentSubID := "sub-parent-inh"
+		s.NoError(s.GetStores().SubscriptionRepo.Create(s.ctx, baseSub(parentSubID, "parent-i", types.SubscriptionTypeParent, types.SubscriptionStatusActive)))
+		inh := baseSub("sub-inh", "child-i", types.SubscriptionTypeInherited, types.SubscriptionStatusActive)
+		inh.ParentSubscriptionID = lo.ToPtr(parentSubID)
+		s.NoError(s.GetStores().SubscriptionRepo.Create(s.ctx, inh))
+
+		_, err := s.service.UpdateCustomer(s.ctx, "child-i", dto.UpdateCustomerRequest{
+			ParentCustomerID: lo.ToPtr("parent-i"),
+		})
+		s.NoError(err)
```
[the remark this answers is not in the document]
omkar273: The TestUpdateCustomer_ParentCustomerID test was removed entirely in this PR as part of the deprecated ParentCustomerID cleanup — there is no longer a parent customer ID to assert on. The underlying UpdateCustomer behavior is covered by the remaining update tests.

internal/service/feature_usage_tracking.go
```diff
@@ -970,6 +997,16 @@ func (s *featureUsageTrackingService) prepareProcessedEventsV2(ctx context.Conte
...
 
+	// STEP 4a: Hierarchy fallback – if no direct line items, check INHERITED subs
+	if len(lineItems) == 0 {
+		lineItems, err = s.resolveHierarchyLineItems(ctx, customer.ID, meterIDs, &event.Timestamp)
+		if err != nil {
+			s.Logger.Errorw("failed hierarchy line item resolution",
+				"error", err, "event_id", event.ID, "customer_id", customer.ID)
+			return results, err
+		}
+	}
```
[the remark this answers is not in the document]
omkar273: Acknowledged — will switch to ErrorwCtx for consistent context propagation in feature_usage_tracking.go. Minor logging consistency fix tracked for follow-up.

internal/testutil/inmemory_customer_store.go
```diff
@@ -110,6 +118,68 @@ func (s *InMemoryCustomerStore) List(ctx context.Context, filter *types.Customer
...
+
+	parentSubIDs := make([]string, 0, len(parentSubs))
+	for _, parentSub := range parentSubs {
+		parentSubIDs = append(parentSubIDs, parentSub.ID)
+	}
+
+	inheritedFilter := types.NewNoLimitSubscriptionFilter()
+	inheritedFilter.ParentSubscriptionIDs = parentSubIDs
+	inheritedFilter.SubscriptionTypes = []types.SubscriptionType{types.SubscriptionTypeInherited}
+	inheritedSubs, err := s.subscriptionStore.List(ctx, inheritedFilter)
```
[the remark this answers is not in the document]
omkar273: Acknowledged — the in-memory store's ListChildrenFromInheritedSubscriptions doesn't filter by subscription status the same way the production query does (which joins on active/published rows). This is an accepted limitation of the test double for this iteration. Production correctness is covered by the actual ClickHouse/Postgres queries.

internal/service/feature_usage_tracking.go
```diff
@@ -1730,14 +1767,91 @@ func (s *featureUsageTrackingService) GetDetailedUsageAnalytics(ctx context.Cont
...
+
+	// 5. If all customers failed, return an empty response — no panic on nil aggregatedData.
+	if aggregatedData == nil {
+		return &dto.GetUsageAnalyticsResponse{
+			TotalCost: decimal.Zero,
+			Currency:  "",
+			Items:     []dto.UsageAnalyticItem{},
+		}, nil
```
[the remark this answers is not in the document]
omkar273: Acknowledged — if all fetchAnalyticsData calls fail, returning an empty success is misleading. Will surface the underlying error in that case. Tracked as a follow-up hardening item.

internal/service/feature_usage_tracking.go
```diff
@@ -2056,6 +2211,45 @@ func (s *featureUsageTrackingService) fetchSubscriptions(ctx context.Context, cu
...
+				parentSubResp, pFetchErr := subscriptionService.GetSubscription(ctx, parentSubID)
+				if pFetchErr != nil {
+					s.Logger.WarnwCtx(ctx, "failed to fetch parent subscription for analytics",
+						"error", pFetchErr,
+						"parent_subscription_id", parentSubID,
+					)
+					continue
+				}
+				subscriptions = append(subscriptions, parentSubResp.Subscription)
+			}
```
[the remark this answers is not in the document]
omkar273: This is the intentional fix for the child analytics cost = 0 bug. The parent subscription is appended to provide its line items (prices) for cost calculation only — the actual usage events are still fetched using the child's customer ID. The parent sub itself has no usage events under the child's external ID, so no synthetic rows are fabricated. The cost calculator reads line items from whichever subscription has them.

docs/superpowers/plans/2026-03-25-subscription-inheritance-execute.md
```diff
@@ -0,0 +1,684 @@
...
+```
+
+- [ ] **Step 4: Check compile**
+
+```bash
+cd /Users/omkar/Developer/source-code/flexprice/flexprice && go build ./internal/api/...
+```
+
+Expected: errors in `internal/service/` because flat fields are no longer on the struct. `internal/api/` should compile cleanly.
+
```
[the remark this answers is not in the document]
omkar273: Acknowledged — go build does not compile test files. The plan docs are internal AI-generated planning artefacts. The actual build verification uses go build ./... (non-test) and go test ./... separately, which correctly exercises test compilation.

docs/superpowers/plans/2026-03-25-subscription-inheritance-execute.md
```diff
@@ -0,0 +1,684 @@
...
+// Previously-cancelled children are NOT skipped — they get a fresh inherited subscription.
+func (s *subscriptionService) ExecuteSubscriptionInheritance(
+	ctx context.Context,
+	parentSubID string,
+	req *dto.ExecuteSubscriptionInheritanceRequest,
+) (*dto.SubscriptionResponse, error) {
+	// 1. Validate request
+	if err := req.Validate(); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
omkar273: The handler always constructs the req via ShouldBindJSON before passing it to the service, so nil is not reachable from the HTTP path. For direct service callers, req.Validate() is the first call so a nil pointer would panic before any logic runs. Will add an explicit nil guard as a defensive measure — tracked as a minor hardening follow-up.

internal/api/dto/subscription.go
```diff
@@ -252,6 +252,42 @@ func (r *SubscriptionCouponRequest) Validate() error {
...
+			WithHint("Send either internal or external customer IDs for inheritance, not both").
+			Mark(ierr.ErrValidation)
+	}
+	if len(r.CustomerIDsToInheritSubscription) == 0 && len(r.ExternalCustomerIDsToInheritSubscription) == 0 {
+		return ierr.NewError("at least one customer ID is required").
+			WithHint("Provide customer_ids_to_inherit_subscription or external_customer_ids_to_inherit_subscription").
+			Mark(ierr.ErrValidation)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
omkar273: Acknowledged — empty-string IDs (e.g. ['']) pass length validation but are invalid identifiers. Will add a whitespace/empty check in ExecuteSubscriptionInheritanceRequest.Validate() as a follow-up.

internal/api/v1/subscription.go
```diff
@@ -651,3 +651,40 @@ func (h *SubscriptionHandler) TriggerSubscriptionWorkflow(c *gin.Context) {
...
+// @Accept json
+// @Produce json
+// @Security ApiKeyAuth
+// @Param id path string true "Parent subscription ID"
+// @Param request body dto.ExecuteSubscriptionInheritanceRequest true "Inheritance request"
+// @Success 200 {object} dto.SubscriptionResponse
+// @Failure 400 {object} ierr.ErrorResponse "Invalid request"
+// @Failure 404 {object} ierr.ErrorResponse "Subscription not found"
+// @Failure 500 {object} ierr.ErrorResponse "Server error"
+// @Router /subscriptions/{id}/inheritance/execute [post]
```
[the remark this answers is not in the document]
omkar273: Acknowledged — will add the @x-scope annotation to the ExecuteSubscriptionInheritance handler in a follow-up swagger annotation pass.

internal/api/v1/subscription.go
```diff
@@ -651,3 +651,40 @@ func (h *SubscriptionHandler) TriggerSubscriptionWorkflow(c *gin.Context) {
...
+		return
+	}
+	var req dto.ExecuteSubscriptionInheritanceRequest
+	if err := c.ShouldBindJSON(&req); err != nil {
+		c.Error(ierr.WithError(err).
+			WithHint("Invalid request format").
+			Mark(ierr.ErrValidation))
+		return
+	}
+	resp, err := h.service.ExecuteSubscriptionInheritance(c.Request.Context(), id, &req)
```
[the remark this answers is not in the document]
omkar273: The service's ExecuteSubscriptionInheritance calls req.Validate() as its first action, so validation is enforced at the service boundary regardless of whether the handler calls it separately. That said, adding an explicit handler-level Validate() call is cleaner and will be added in a follow-up.

internal/service/subscription.go
```diff
@@ -6150,3 +6239,472 @@ func (s *subscriptionService) TriggerSubscriptionWorkflow(ctx context.Context, s
...
+) error {
+	usageCustomerIDs = lo.Uniq(usageCustomerIDs)
+	allCustomerIDs := lo.Uniq(append([]string{parentCustomerID}, usageCustomerIDs...))
+	filter := types.NewNoLimitSubscriptionFilter()
+	filter.CustomerIDs = allCustomerIDs
+	filter.SubscriptionStatus = []types.SubscriptionStatus{
+		types.SubscriptionStatusActive,
+		types.SubscriptionStatusTrialing,
+		types.SubscriptionStatusDraft,
+	}
```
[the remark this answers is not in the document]
omkar273: Valid — paused standalone/parent subscriptions should block inheritance mixing. The current filter only checks active/trialing/draft. Will extend validateCustomerSubscriptionWorkflow to include paused subscriptions in the conflict check.

internal/service/subscription.go
```diff
@@ -6150,3 +6239,472 @@ func (s *subscriptionService) TriggerSubscriptionWorkflow(ctx context.Context, s
...
+
+	// 8. Validate no STANDALONE conflicts for the filtered set
+	if err := s.validateCustomerSubscriptionWorkflow(ctx, parentSub.CustomerID, filtered); err != nil {
+		return nil, err
+	}
+
+	// 9. Create inherited subscriptions
+	if err := s.createInheritedSubscriptions(ctx, parentSub, filtered); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
omkar273: Valid concern — createInheritedSubscriptions should be wrapped in a DB transaction so partial failures don't leave orphaned inherited rows. Will wrap in s.DB.WithTx() as a follow-up hardening item.

docs/swagger/swagger.json
```diff
@@ -13744,17 +13809,9 @@
...
-                    "description": "invoicing_customer_external_id is the external ID of the customer to use for invoicing.\nResolved internally to an internal customer ID via external ID lookup.\nMutually exclusive with invoicing_customer_id.",
-                    "type": "string"
-                },
-                "invoicing_customer_id": {
-                    "description": "invoicing_customer_id is the FlexPrice customer ID to use for invoicing.\nThis can differ from the subscription customer (e.g., a billing entity invoicing on behalf of another customer).\nMutually exclusive with invoicing_customer_external_id.",
-                    "type": "string"
+                "inheritance": {
+                    "description": "Inheritance groups all customer-hierarchy fields.\nWhen provided with at least one child ID, the subscription becomes a PARENT type.",
+                    "$ref": "#/definitions/dto.SubscriptionInheritanceConfig"
                 },
```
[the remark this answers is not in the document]
omkar273: Acknowledged — swagger docs for the new inheritance endpoints will be reviewed and made symmetric in a follow-up swagger cleanup pass.

docs/superpowers/specs/2026-03-25-remove-deprecated-fields-design.md
```diff
@@ -0,0 +1,179 @@
...
+From `CreateCustomerRequest`, delete:
+```go
+ParentCustomerID         *string `json:"parent_customer_id,omitempty"`
+ParentCustomerExternalID *string `json:"parent_customer_external_id,omitempty"`
+```
+
+From `UpdateCustomerRequest`, delete the same two fields.
+
+From `CustomerResponse`, delete:
+```go
```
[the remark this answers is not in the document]
omkar273: Acknowledged — minor wording issue in the internal spec doc. Will clean up in a follow-up docs pass.

internal/api/dto/events.go
```diff
@@ -316,16 +319,26 @@ func (r *GetEventsRequest) Validate() error {
...
-	WindowSize         types.WindowSize `json:"window_size,omitempty"`
-	Expand             []string         `json:"expand,omitempty"` // allowed values: "price", "meter", "feature", "subscription_line_item","plan","addon"
+	// ExternalCustomerID is the single external customer ID.
+	// Optional when ExternalCustomerIDs is provided; required otherwise.
+	ExternalCustomerID string `json:"external_customer_id"`
+	// ExternalCustomerIDs is a list of external customer IDs whose usage will be merged
+	// into a single aggregated response. Unioned with ExternalCustomerID if both are set;
+	// duplicates are dropped. At least one of ExternalCustomerID or ExternalCustomerIDs
+	// must be provided.
+	ExternalCustomerIDs []string         `json:"external_customer_ids,omitempty"`
```
nkmishra1997, line 329: remove json tags on all these external_customer_ids fields because this can be abused a lot

internal/api/dto/subscription.go
```diff
@@ -252,6 +252,41 @@ func (r *SubscriptionCouponRequest) Validate() error {
...
 
+// SubscriptionInheritanceConfig groups all inheritance-related fields for
+// subscription creation. InvoicingCustomerID and InvoicingCustomerExternalID
+// use pointer semantics so nil ("not provided") differs from "" ("explicitly empty").
+type SubscriptionInheritanceConfig struct {
+	CustomerIDsToInheritSubscription         []string              `json:"customer_ids_to_inherit_subscription,omitempty"`
+	ExternalCustomerIDsToInheritSubscription []string              `json:"external_customer_ids_to_inherit_subscription,omitempty"`
+	ParentSubscriptionID                     string                `json:"parent_subscription_id,omitempty"`
+	InvoicingCustomerID                      *string               `json:"invoicing_customer_id,omitempty"`
+	InvoicingCustomerExternalID              *string               `json:"invoicing_customer_external_id,omitempty"`
```
nkmishra1997, line 263: we should remove customer_ids flags from DTOs (can use with json -) and always use external ids from external world

internal/api/dto/subscription.go
```diff
@@ -364,14 +383,18 @@ type CreateSubscriptionRequest struct {
...
 
 	// Enable Commitment True Up Fee
 	EnableTrueUp bool `json:"enable_true_up"`
+
+	// Inheritance groups all customer-hierarchy fields.
+	// When provided with at least one child ID, the subscription becomes a PARENT type.
+	Inheritance *SubscriptionInheritanceConfig `json:"inheritance,omitempty"`
+
+	// SubscriptionType is set internally by the service layer.
+	SubscriptionType types.SubscriptionType `json:"-"`
```
nkmishra1997, line 397: can we give default value here itself.

internal/api/v1/subscription.go
nkmishra1997, line 668: this can not be part of docs - we need a single update API

internal/api/router.go
```diff
@@ -297,6 +297,9 @@ func NewRouter(handlers Handlers, cfg *config.Configuration, logger *logger.Logg
 			subscription.POST("/:id/change/preview", handlers.SubscriptionChange.PreviewSubscriptionChange)
 			subscription.POST("/:id/change/execute", handlers.SubscriptionChange.ExecuteSubscriptionChange)
 
+			// Subscription inheritance management
+			subscription.POST("/:id/inheritance/execute", handlers.Subscription.ExecuteSubscriptionInheritance)
```
nkmishra1997, line 301: remove in followup

subscription.POST("/:id/update/execute")

internal/domain/events/repository.go
```diff
@@ -82,17 +85,18 @@ type UsageAnalytic struct {
...
-	PropertyName       string                `json:"property_name" validate:"required"`
-	AggregationType    types.AggregationType `json:"aggregation_type" validate:"required"`
-	WindowSize         types.WindowSize      `json:"window_size"`
-	BucketSize         types.WindowSize      `json:"bucket_size,omitempty"` // For windowed MAX aggregation
-	StartTime          time.Time             `json:"start_time" validate:"required"`
-	EndTime            time.Time             `json:"end_time" validate:"required"`
-	Filters            map[string][]string   `json:"filters"`
-	Multiplier         *decimal.Decimal      `json:"multiplier,omitempty" validate:"omitempty,gt=0"`
+	ExternalCustomerID  string                `json:"external_customer_id"`
+	ExternalCustomerIDs []string              `json:"external_customer_ids"`
```
nkmishra1997, line 89: remove json tag

internal/repository/clickhouse/aggregators.go
nkmishra1997, line 268: follow up: move this also inside buildExternalCustomerFilter -> buildCustomerFilter

internal/repository/ent/customer.go
```diff
@@ -242,6 +241,94 @@ func (r *customerRepository) List(ctx context.Context, filter *types.CustomerFil
...
+		INNER JOIN subscriptions AS child_sub
+			ON child_sub.parent_subscription_id = parent_sub.id
+			AND child_sub.tenant_id = parent_sub.tenant_id
+			AND child_sub.environment_id = parent_sub.environment_id
+		WHERE parent_sub.customer_id = $1
+			AND parent_sub.subscription_type = $2
+			AND child_sub.subscription_type = $3
+			AND parent_sub.tenant_id = $4
+			AND parent_sub.environment_id = $5
+			AND child_sub.customer_id <> $1
```
nkmishra1997, line 267: i think we should break this in 2 queries:
-- Query 1: get 1-2 rows max
SELECT id FROM subscriptions 
WHERE customer_id = $1 AND tenant_id = $2 AND environment_id = $3 
AND subscription_type = $4

-- Query 2: get child customer IDs
SELECT DISTINCT customer_id FROM subscriptions
WHERE parent_subscription_id IN ($id1, $id2)
AND tenant_id = $2 AND environment_id = $3
AND subscription_type = $5
AND customer_id <> $1
```

Both are point lookups. Both will be index scans returning tiny row counts. The network round trip is genuinely the only real cost here.

File: internal/service/event_post_processing.go [hunk 58]
```diff
@@ -953,6 +953,17 @@ func (s *eventPostProcessingService) GetUsageAnalytics(ctx context.Context, tena
...
+
```
nkmishra1997, line 966: simplify

internal/service/feature_usage_tracking.go
```diff
@@ -562,12 +562,39 @@ func (s *featureUsageTrackingService) prepareProcessedEvents(ctx context.Context
...
+				types.SubscriptionStatusActive,
+				types.SubscriptionStatusTrialing,
+			}
+			parentList, pErr := subscriptionService.ListSubscriptions(ctx, parentFilter)
+			if pErr != nil {
+				return results, pErr
+			}
+			subscriptions = parentList.Items
+		}
+	}
```
nkmishra1997, line 591: this is invalid

internal/service/feature_usage_tracking.go
```diff
@@ -562,12 +562,39 @@ func (s *featureUsageTrackingService) prepareProcessedEvents(ctx context.Context
...
+		}
+		if len(parentSubIDs) > 0 {
+			parentFilter := types.NewSubscriptionFilter()
+			parentFilter.SubscriptionIDs = parentSubIDs
+			parentFilter.WithLineItems = true
+			parentFilter.Expand = lo.ToPtr(string(types.ExpandPrices) + "," + string(types.ExpandMeters))
+			parentFilter.SubscriptionStatus = []types.SubscriptionStatus{
+				types.SubscriptionStatusActive,
+				types.SubscriptionStatusTrialing,
+			}
```
nkmishra1997, line 584: we can also remove `h.featureUsageTrackingService.BenchmarkPrepareV1(ctx, event)`

internal/service/subscription.go
```diff
@@ -6150,3 +6235,472 @@ func (s *subscriptionService) TriggerSubscriptionWorkflow(ctx context.Context, s
...
+			resolvedIDs = append(resolvedIDs, child.ID)
+		}
+		return lo.Uniq(resolvedIDs), nil
+	}
+	return nil, nil
+}
+
+// validateCustomerSubscriptionWorkflow ensures a customer is not mixing STANDALONE
+// subscriptions with hierarchy-based subscriptions (PARENT or INHERITED).
+func (s *subscriptionService) validateCustomerSubscriptionWorkflow(
```
nkmishra1997, line 6309: can rename this

internal/service/subscription.go
```diff
@@ -6150,3 +6235,472 @@ func (s *subscriptionService) TriggerSubscriptionWorkflow(ctx context.Context, s
...
 }
+
+// resolveUsageCustomerIDs determines the set of customer IDs whose usage
+// should be aggregated on the subscription.
+//   - If req.Inheritance is nil, the subscription is STANDALONE.
+//   - If req.Inheritance.CustomerIDsToInheritSubscription is non-empty, use those IDs.
+//   - Else if req.Inheritance.ExternalCustomerIDsToInheritSubscription is non-empty, resolve them to internal IDs.
+//   - Otherwise (both arrays absent/empty), treat as STANDALONE.
+//   - Returns nil/empty when the subscription should be STANDALONE.
+func (s *subscriptionService) resolveUsageCustomerIDs(
```
nkmishra1997, line 6246: rename methods

internal/service/subscription.go
```diff
@@ -6150,3 +6235,472 @@ func (s *subscriptionService) TriggerSubscriptionWorkflow(ctx context.Context, s
...
+		}
+		externalIDs = append(externalIDs, ext)
+	}
+
+	return externalIDs, nil
+}
+
+// getAggregatedCustomerIDs returns external customer IDs to aggregate usage for a subscription.
+// PARENT type includes the subscription customer plus all hierarchy children; other types only that customer.
+func (s *subscriptionService) getAggregatedCustomerIDs(ctx context.Context, sub *subscription.Subscription) ([]string, error) {
```
nkmishra1997, line 6409:
```suggestion
func (s *subscriptionService) getAggregatedExternalCustomerIDs(ctx context.Context, sub *subscription.Subscription) ([]string, error) {
```

internal/service/feature_usage_tracking.go
```diff
@@ -2056,6 +2211,45 @@ func (s *featureUsageTrackingService) fetchSubscriptions(ctx context.Context, cu
...
+		if pErr != nil {
+			s.Logger.WarnwCtx(ctx, "failed to fetch parent subscription IDs for inherited customer; costs may be $0",
+				"error", pErr,
+				"customer_id", customerID,
+			)
+		} else {
+			seen := make(map[string]bool, len(subscriptions))
+			for _, sub := range subscriptions {
+				seen[sub.ID] = true
+			}
```
nkmishra1997, line 2234: this will never get used

internal/service/feature_usage_tracking.go
```diff
@@ -2056,6 +2211,45 @@ func (s *featureUsageTrackingService) fetchSubscriptions(ctx context.Context, cu
...
+						"error", pFetchErr,
+						"parent_subscription_id", parentSubID,
+					)
+					continue
+				}
+				subscriptions = append(subscriptions, parentSubResp.Subscription)
+			}
+		}
+	}
+
```
nkmishra1997, line 2252: review whole thing can be super optimised

internal/service/feature_usage_tracking.go
```diff
@@ -3595,21 +3789,29 @@ func (s *featureUsageTrackingService) fetchAddons(ctx context.Context, data *Ana
...
 		return customers, nil
 	}
+	// No IDs specified — fetch all customers (existing behaviour for V2 aggregate-all path)
+	customers, err := s.CustomerRepo.List(ctx, types.NewNoLimitCustomerFilter())
+	if err != nil {
+		return nil, ierr.WithError(err).
+			WithHint("Failed to fetch customers").
+			Mark(ierr.ErrDatabase)
+	}
+	return customers, nil
```
nkmishra1997, line 3814: this is fetching all customers - lets just remove this

internal/service/feature_usage_tracking.go
```diff
@@ -3595,21 +3789,29 @@ func (s *featureUsageTrackingService) fetchAddons(ctx context.Context, data *Ana
...
 }
```
nkmishra1997, line 3816: in gereral - remove the case in codebase where we are fetching all customers for cost analytics

internal/service/feature_usage_tracking.go
```diff
@@ -3595,21 +3789,29 @@ func (s *featureUsageTrackingService) fetchAddons(ctx context.Context, data *Ana
...
-		cust, err := s.fetchCustomer(ctx, req.ExternalCustomerID)
-		if err != nil {
-			return nil, err
-		}
-		return []*customer.Customer{cust}, nil
-	} else {
-		customers, err := s.CustomerRepo.List(ctx, types.NewNoLimitCustomerFilter())
-		if err != nil {
-			return nil, ierr.WithError(err).
-				WithHint("Failed to fetch customers").
```
nkmishra1997: remove all code where we are fetching all customers

PR #1420, opened 2026-03-20
feat(subscription): added handling of inactive line item for previous…

hiteshshimpi-55:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed boundary condition so subscription line items that end exactly at a reference period are included when appropriate.

* **New Features**
  * Historical usage and billing queries now reliably include relevant past line items for accurate charge and usage calculations.

* **Tests**
  * Expanded and updated tests and in-memory test stores to validate historical line-item retrieval, ordering, and the updated inclusion behavior.

### DIFF
--- a/internal/service/subscription.go
+++ b/internal/service/subscription.go
@@ -2035,7 +2035,7 @@ func (s *subscriptionService) GetUsageBySubscription(ctx context.Context, req *d
 	priceService := NewPriceService(s.ServiceParams)
 
 	// Get subscription with line items
-	subscription, lineItems, err := s.SubRepo.GetWithLineItems(ctx, req.SubscriptionID)
+	subscription, err := s.SubRepo.Get(ctx, req.SubscriptionID)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/service/subscription.go
nkmishra1997: make this simple subscription get without the line items

internal/service/subscription.go
```diff
@@ -2045,6 +2062,11 @@ func (s *subscriptionService) GetUsageBySubscription(ctx context.Context, req *d
 		usageEndTime = time.Now().UTC()
 	}
 
+	lineItems, err = s.listSubscriptionLineItemsForUsageWindow(ctx, subscription.ID, usageStartTime, req.LifetimeUsage)
```
nkmishra1997: then this should append line items in the subscription as well

internal/service/subscription.go
nkmishra1997: same here

PR #1441, opened 2026-03-25
fix(billing): cap periodEnd to now() in preview and wallet balance paths

hiteshshimpi-55:
## Summary by CodeRabbit

* **Bug Fixes**
  * Invoice preview calculations now correctly limit commitment and true-up charges to elapsed time, capping the billing period to the current moment instead of extending into future dates. This ensures accurate preview amounts.
  * Wallet balance calculations across all methods now similarly restrict charges to elapsed time, providing consistent and accurate balance reporting.

### DIFF
--- a/internal/service/invoice.go
+++ b/internal/service/invoice.go
@@ -1787,6 +1787,13 @@ func (s *invoiceService) GetPreviewInvoice(ctx context.Context, req dto.GetPrevi
 		req.PeriodEnd = &sub.CurrentPeriodEnd
 	}
 
+	// Cap periodEnd to now() so commitment/true-up is only calculated for elapsed time
+	now := time.Now()
+	if req.PeriodEnd.After(now) {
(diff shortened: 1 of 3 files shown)


internal/service/invoice.go
```diff
@@ -1787,6 +1787,13 @@ func (s *invoiceService) GetPreviewInvoice(ctx context.Context, req dto.GetPrevi
 		req.PeriodEnd = &sub.CurrentPeriodEnd
 	}
 
+	// Cap periodEnd to now() so commitment/true-up is only calculated for elapsed time
+	now := time.Now()
+	if req.PeriodEnd.After(now) {
+		cappedEnd := now
+		req.PeriodEnd = &cappedEnd
+	}
+
```
nkmishra1997, line 1796: remove this since we are already clamping the end date inside the billing logic

PR #1505, opened 2026-03-31
feat(invoice): refactored finalization due logic

hiteshshimpi-55:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed invoice finalization timing for subscription-cycle invoices. If a prior computation occurred before the billing period ended, finalization will now be deferred so invoices aren't finalized prematurely. This ensures the configured finalization delay is respected and invoices are processed at the correct time, reducing incorrect or early invoice closures.

### DIFF
--- a/internal/service/invoice.go
+++ b/internal/service/invoice.go
@@ -969,7 +969,7 @@ func (s *invoiceService) IsFinalizationDue(ctx context.Context, invoiceID string
 		return false, nil
 	}
 
-	if inv.LastComputedAt != nil && inv.LastComputedAt.Before(*inv.PeriodEnd) {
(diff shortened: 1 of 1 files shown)


internal/service/invoice.go
```diff
@@ -969,6 +969,10 @@ func (s *invoiceService) IsFinalizationDue(ctx context.Context, invoiceID string
 		return false, nil
 	}
 
+	if inv.LastComputedAt != nil && inv.LastComputedAt.Before(*inv.PeriodEnd) {
```
nkmishra1997: billing reason check should also be there

nkmishra1997 (approved):

PR #1573, opened 2026-04-09
perf(meter-usage): cache meter list lookups per event name

nkmishra1997:
## Changes

- Add in-process caching layer for meter list lookups using go-cache
- Cache key: `tenantID:environmentID:eventName` with 10-minute TTL
- Introduce `getMetersForEvent()` helper method to encapsulate cache logic
- Add comprehensive documentation explaining cache design decisions

## Benefits

- Eliminates redundant Postgres queries for the same event name within the TTL window
- Improves Kafka message processing latency on high-throughput topics
- Negligible memory footprint (~20 MB worst case for 200 tenants)
- Safe middle ground: picks up new meters within 10 minutes without requiring process restart

## Implementation Details

- Uses dedicated in-process cache (not Redis) for guaranteed sub-millisecond lookup
- Meters are immutable after creation, so no active invalidation needed
- Cache is nil-safe for backward compatibility with tests
- Singleton cache instance per process via fx dependency injection

## Summary by CodeRabbit

* **Refactor**
  * Added an internal in-process cache for meter lookups to reduce redundant queries during event processing and improve performance with no public API changes.

* **Bug Fix**
  * Improved deduplication hash generation so unique counts are consistently computed even when expected event fields are missing, preventing empty-hash cases.

### DIFF
--- a/internal/service/meter_usage_tracking.go
+++ b/internal/service/meter_usage_tracking.go
@@ -19,9 +19,34 @@ import (
 	"github.com/flexprice/flexprice/internal/pubsub/kafka"
 	pubsubRouter "github.com/flexprice/flexprice/internal/pubsub/router"
 	"github.com/flexprice/flexprice/internal/types"
+	goCache "github.com/patrickmn/go-cache"
 	"github.com/samber/lo"
 	"github.com/shopspring/decimal"
 )
+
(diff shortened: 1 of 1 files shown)


nkmishra1997:
@coderabbitai please review

hiteshshimpi-55 (approved):

PR #1609, opened 2026-04-15
fix(webhooks): write to system events immediately

nkmishra1997:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved webhook error logging and changed delivery failure handling so delivery errors are logged but no longer cause automatic retries or propagate as failures.
* **Refactor**
  * Webhook publishing now records message consumption state before sending; failures to record are logged but do not block delivery. Internal wiring updated to support this behavior.

### DIFF
--- a/internal/testutil/base_service_suite.go
+++ b/internal/testutil/base_service_suite.go
@@ -141,7 +141,7 @@ func (s *BaseServiceTestSuite) setupDependencies() {
 	eventStore := s.stores.EventRepo.(*InMemoryEventStore)
 	s.publisher = NewInMemoryEventPublisher(eventStore)
 	pubsub := NewInMemoryPubSub()
-	webhookPublisher, err := webhookPublisher.NewPublisher(pubsub, s.config, s.logger)
+	webhookPublisher, err := webhookPublisher.NewPublisher(pubsub, s.config, s.logger, nil)
 	if err != nil {
 		s.T().Fatalf("failed to create webhook publisher: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


internal/webhook/handler/handler.go
```diff
@@ -145,21 +145,30 @@ func (h *handler) processMessage(msg *message.Message) error {
...
-			"event_name", event.EventName,
-		)
-	}
-
 	// After verify: deliver the webhook (Svix or native endpoint)
 	if h.config.Svix.Enabled {
-		return h.processMessageSvix(ctx, &event, msg.UUID)
+		err := h.processMessageSvix(ctx, &event, msg.UUID)
+		if err != nil {
+			h.logger.Errorw("failed to send webhook via Svix",
```
subratsahilgupta, line 152: not needed

internal/webhook/publisher/publisher.go
```diff
@@ -84,6 +90,14 @@ func (p *webhookPublisher) PublishWebhook(ctx context.Context, event *types.Webh
 		"payload", string(payload),
 	)
 
+	if err := p.systemEventRepo.OnConsumed(ctx, event); err != nil {
+		p.logger.ErrorwCtx(ctx, "system_events OnConsumed failed",
+			"error", err,
+			"event_id", event.ID,
+			"event_name", event.EventName,
+		)
+	}
```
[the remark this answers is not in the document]
subratsahilgupta: not optional, support would be added later

internal/testutil/base_service_suite.go
```diff
@@ -141,7 +141,7 @@ func (s *BaseServiceTestSuite) setupDependencies() {
 	eventStore := s.stores.EventRepo.(*InMemoryEventStore)
 	s.publisher = NewInMemoryEventPublisher(eventStore)
 	pubsub := NewInMemoryPubSub()
-	webhookPublisher, err := webhookPublisher.NewPublisher(pubsub, s.config, s.logger)
+	webhookPublisher, err := webhookPublisher.NewPublisher(pubsub, s.config, s.logger, nil)
```
subratsahilgupta, line 144: support to be added later in follow up

subratsahilgupta (approved):

internal/webhook/handler/handler.go
```diff
@@ -145,21 +145,30 @@ func (h *handler) processMessage(msg *message.Message) error {
...
+				"error", err,
+				"message_uuid", msg.UUID,
+				"tenant_id", event.TenantID,
+				"event", event.EventName,
+			)
+			return nil
+		}
 	}
```
subratsahilgupta, line 160: return nil here instead of early

subratsahilgupta: @nkmishra1997

PR #1618, opened 2026-04-17
fix(environments): fetch only active subscriptions

subratsahilgupta:
## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced that using a billing anchor requires Anniversary billing cycle.
  * Environments listing now returns only published environments and correct total counts.

* **Changes**
  * Removed time-based validation comparing billing anchor against start date; billing-anchor validity is now gated by billing-cycle type.

* **Tests**
  * Removed a test that validated time-based billing-anchor vs start-date behavior.


internal/api/dto/subscription.go
```diff
@@ -637,17 +634,14 @@ func (r *CreateSubscriptionRequest) Validate() error {
...
-	if r.BillingAnchor != nil {
-		if r.BillingAnchor.Before(lo.FromPtr(r.StartDate)) {
-			return ierr.NewError("billing_anchor cannot be before start_date").
-				WithHint("billing_anchor must be on or after start_date").
-				WithReportableDetails(map[string]any{
-					"billing_anchor": r.BillingAnchor,
-					"start_date":     r.StartDate,
-				}).
-				Mark(ierr.ErrValidation)
-		}
```
omkar273, line 650: Why are we removing this check?

internal/service/environment.go
```diff
@@ -143,16 +143,20 @@ func (s *environmentService) GetEnvironments(ctx context.Context, filter types.F
...
-		Total:        len(allowedEnvironments),
+		Environments: make([]dto.EnvironmentResponse, 0, len(allowedEnvironments)),
+		Total:        0,
 		Offset:       filter.Offset,
 		Limit:        filter.Limit,
 	}
 
-	for i, env := range allowedEnvironments {
-		response.Environments[i] = *dto.NewEnvironmentResponse(env)
+	for _, env := range allowedEnvironments {
```
omkar273, line 152: Instead of this, use the `status` field in `types.Filter` and update the calling sites to pass `status=published`

omkar273: This will only fetch only published environments from db so no need of filtering unecessary

internal/api/dto/subscription_test.go
```diff
@@ -51,62 +51,3 @@ func TestCreateSubscriptionRequestValidate_BillingAnchorRequiresAnniversaryBilli
...
-		req.BillingCycle = types.BillingCycleAnniversary
-		anchor := time.Date(2024, 1, 15, 0, 0, 0, 0, time.UTC)
-		req.BillingAnchor = &anchor
-
-		err := req.Validate()
-		if err != nil {
-			t.Fatalf("expected no error, got: %v", err)
-		}
-	})
-}
```
omkar273, line 112: Lets revert this change

PR #1622, opened 2026-04-20
fix(stripe): always set due_date for send_invoice

nkmishra1997:
## Summary
- Stripe rejects `send_invoice` creation when neither `due_date` nor `days_until_due` is provided
- Previously only set `due_date` when `flexInvoice.DueDate` was non-nil — invoices without a due date would fail
- Now: use invoice due date if valid (≥15 min from now), otherwise fall back to `now + InvoiceDefaultDueDays`

## Test plan
- [ ] Sync an invoice with `collection_method=send_invoice` and a valid due date — uses the invoice due date
- [ ] Sync an invoice with `collection_method=send_invoice` and no due date — uses default (now + 1 day)
- [ ] Sync an invoice with `collection_method=send_invoice` and a past/stale due date — uses default

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved invoice due date handling in Stripe payment collection. Invoices now consistently have a due date assigned, with a default 15-minute grace period applied when no custom due date is specified.

### DIFF
--- a/internal/integration/stripe/invoice_sync.go
+++ b/internal/integration/stripe/invoice_sync.go
@@ -171,14 +171,13 @@ func (s *InvoiceSyncService) createDraftInvoiceInStripe(ctx context.Context, fle
 		params.CollectionMethod = stripe.String(string(stripe.InvoiceCollectionMethodChargeAutomatically))
 	}
 
-	// Set due date only for send_invoice collection method.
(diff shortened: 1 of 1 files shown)


omkar273:
loogs good to me

omkar273 (approved):

PR #1636, opened 2026-04-22
fix(onboarding-user): optimize onboarding events and invite user workflows

subratsahilgupta:
## Summary by CodeRabbit

* **Refactor**
  * Simplified error handling in authentication and invitation flows — errors are now propagated with less wrapping for clearer reporting.
  * Onboarding now initializes its event pubsub internally (Kafka) and logs when event publishing is unavailable instead of requiring external injection.
  * Standardized user record initialization across service operations for consistent metadata.

* **Tests**
  * Updated test wiring to use the onboarding service dependency instead of an injected pubsub.

### DIFF
--- a/internal/service/auth.go
+++ b/internal/service/auth.go
@@ -86,7 +86,7 @@ func (s *authService) SignUp(ctx context.Context, req *dto.SignUpRequest) (*dto.
 			}
 		}
 
-		onboardingService := NewOnboardingService(s.ServiceParams, s.pubSub)
+		onboardingService := NewOnboardingService(s.ServiceParams)
 
 		err = onboardingService.OnboardNewUserWithTenant(
 			ctx,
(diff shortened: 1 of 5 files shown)


internal/service/auth.go
```diff
@@ -18,18 +17,18 @@ type AuthService interface {
...
-	pubSub       pubsub.PubSub
-	authProvider authProvider.Provider
+	onboardingService OnboardingService
+	authProvider      authProvider.Provider
 }
 
 func NewAuthService(
 	params ServiceParams,
-	pubSub pubsub.PubSub,
+	onboardingService OnboardingService,
```
nkmishra1997: why is auth service dependent on onboarding service now?

subratsahilgupta: during signup it is used for assigning the new user to a tenant

internal/service/auth.go
```diff
@@ -86,9 +85,7 @@ func (s *authService) SignUp(ctx context.Context, req *dto.SignUpRequest) (*dto.
 			}
 		}
 
-		onboardingService := NewOnboardingService(s.ServiceParams, s.pubSub)
-
-		err = onboardingService.OnboardNewUserWithTenant(
+		err = s.onboardingService.OnboardNewUserWithTenant(
```
nkmishra1997: this was correct

nkmishra1997 (approved):

PR #1821, opened 2026-05-20
feat(integration): add support for whop invoice sync

subratsahilgupta:
## Summary by CodeRabbit

* **New Features**
  * Whop integration: connect accounts, manage products/plans, create and fetch invoices, provider discovery.
  * Automatic invoice sync to Whop and background workflows/activities to mark invoices paid.
  * Public Whop webhook endpoint that always acknowledges requests and processes invoice.paid events.
  * New configuration option to override Whop base URL.

* **Bug Fixes**
  * Improved startup/shutdown timeouts for more resilient service lifecycle.
  * Enhanced metadata handling and validation for Whop connections.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flexprice/flexprice/pull/1821?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


internal/api/v1/webhook.go
```diff
@@ -1175,3 +1176,60 @@ func (h *WebhookHandler) HandlePaddleWebhook(c *gin.Context) {
 			"event_type", payload.EventType)
 	}
 }
+func (h *WebhookHandler) HandleWhopWebhook(c *gin.Context) {
+	// Always return 200 OK to Whop to prevent retries
+	defer func() {
+		c.JSON(http.StatusOK, gin.H{"message": "ok"})
+	}()
```
omkar273, line 1186: we should follow the consistent pattern across all handlers 

```
handled := false
	defer func() {
		if !handled {
			c.JSON(http.StatusOK, gin.H{
				"message": "Webhook received",
			})
		}
	}()

```

internal/integration/whop/webhook/handler.go
```diff
@@ -0,0 +1,110 @@
...
+// HandleWebhookEvent processes an inbound Whop webhook event.
+// Always returns nil so Whop receives 200 OK.
+func (h *Handler) HandleWebhookEvent(ctx context.Context, event *WhopWebhookEvent, services *ServiceDependencies) error {
+	h.logger.Infow("processing Whop webhook event",
+		"type", event.Type,
+		"whop_invoice_id", event.Data.ID,
+		"msg_id", event.ID)
+
+	switch event.Type {
+	case "invoice.paid":
```
omkar273: we should create enums for webhook event types of whop

internal/integration/whop/client.go
```diff
@@ -0,0 +1,318 @@
...
+	if whopConfig.APIKey == "" {
+		return nil, ierr.NewError("missing Whop API key").
+			WithHint("Configure Whop API key in the connection settings").
+			Mark(ierr.ErrValidation)
+	}
+	if whopConfig.CompanyID == "" {
+		return nil, ierr.NewError("missing Whop company ID").
+			WithHint("Configure Whop company ID in the connection settings").
+			Mark(ierr.ErrValidation)
+	}
```
omkar273, line 89: Evaluate whether we should also add validation for `product_id` here or not

internal/integration/whop/dto.go
```diff
@@ -0,0 +1,85 @@
+package whop
+
+// Constants for Whop integration
+const (
+	// WhopBaseURL is the Whop API base URL. Sandbox for now; swap to api.whop.com for prod.
+	WhopBaseURL = "https://api.whop.com"
+	// WhopBaseURL = "https://sandbox-api.whop.com" // for sandbox whop testing
```
omkar273: We should keep it configurable based on api key since there should be a way to test it on whop sandbox as well

internal/integration/whop/dto.go
```diff
@@ -0,0 +1,85 @@
...
+	EmailAddress     string      `json:"email_address"`
+}
+
+// InvoiceResponse is the response from POST /v1/invoices
+type InvoiceResponse struct {
+	ID          string `json:"id"`
+	Status      string `json:"status"`
+	CurrentPlan struct {
+		ID string `json:"id"`
+	} `json:"current_plan"`
```
omkar273: Create a struct for the current plan separately and use it here instead of inline implementation

internal/integration/whop/dto.go
```diff
@@ -0,0 +1,85 @@
...
+	ID            string                 `json:"id"`
+	PlanType      string                 `json:"plan_type"`
+	PurchaseURL   string                 `json:"purchase_url"`
+	Currency      string                 `json:"currency"`
+	InternalNotes string                 `json:"internal_notes"`
+	Metadata      map[string]interface{} `json:"metadata"`
+	Product       struct {
+		ID    string `json:"id"`
+		Title string `json:"title"`
+	} `json:"product"`
```
omkar273: Create a struct for the product separately and use it here instead of inline implementation

internal/integration/whop/dto.go
```diff
@@ -0,0 +1,85 @@
...
+	Invoice struct {
+		ID string `json:"id"`
+	} `json:"invoice"`
```
omkar273: Create a struct for the invoice separately and use it here instead of inline implementation

internal/integration/whop/dto.go
```diff
@@ -0,0 +1,85 @@
...
+	InitialPrice float64 `json:"initial_price"`
```
omkar273: We should use `decimal.Decimal` instead of `float64` here

internal/integration/whop/dto.go
```diff
@@ -0,0 +1,85 @@
...
+
+// WhopInvoiceSyncRequest is the request for syncing a Flexprice invoice to Whop
+type WhopInvoiceSyncRequest struct {
+	InvoiceID string
+}
+
+// WhopInvoiceSyncResponse is the response from syncing a Flexprice invoice to Whop
+type WhopInvoiceSyncResponse struct {
+	WhopInvoiceID string
+	Status        string
```
omkar273, line 115: Create an enum for possible `status` values of whop

internal/integration/whop/invoice_sync.go
```diff
@@ -0,0 +1,306 @@
...
+
+	if !s.client.HasWhopConnection(ctx) {
+		return nil, ierr.NewError("Whop connection not available").
+			WithHint("Whop integration must be configured for invoice sync").
+			Mark(ierr.ErrNotFound)
+	}
+
+	flexInvoice, err := s.invoiceRepo.Get(ctx, req.InvoiceID)
+	if err != nil {
+		return nil, ierr.WithError(err).WithHint("Failed to get Flexprice invoice").Mark(ierr.ErrDatabase)
```
omkar273: directly return the error without wrapping

internal/integration/whop/invoice_sync.go
```diff
@@ -0,0 +1,306 @@
...
+	if err != nil && !ierr.IsNotFound(err) {
+		return nil, err
+	}
+	if existingMapping != nil {
+		s.logger.Infow("invoice already synced to Whop",
+			"invoice_id", req.InvoiceID,
+			"whop_invoice_id", existingMapping.ProviderEntityID)
+		return &WhopInvoiceSyncResponse{
+			WhopInvoiceID: existingMapping.ProviderEntityID,
+			Status:        "already_synced",
```
omkar273: use enum instead of harcoded values

internal/integration/whop/invoice_sync.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	customerName, customerEmail, err := s.resolveCustomer(ctx, flexInvoice, customerService)
+	if err != nil {
+		return nil, err
+	}
+
+	totalFloat, _ := flexInvoice.AmountDue.Round(2).Float64()
+
+	defaultDueDate := time.Now().AddDate(0, 0, 30).UTC()
```
omkar273: Instead of hardcoding it here, we should define a default invoice due days variable in the config/constants and use that value here.

internal/integration/whop/invoice_sync.go
```diff
@@ -0,0 +1,306 @@
...
+		if planErr != nil {
+			s.logger.Warnw("failed to fetch Whop plan for purchase_url",
+				"plan_id", whopInvoice.CurrentPlan.ID, "error", planErr)
+		} else if plan.PurchaseURL != "" {
+			if updateErr := s.storeCheckoutURL(ctx, req.InvoiceID, plan.PurchaseURL); updateErr != nil {
+				s.logger.Warnw("failed to store Whop purchase_url on invoice",
+					"invoice_id", req.InvoiceID, "error", updateErr)
+			}
+		}
+	}
```
omkar273, line 137: We can simplify this by removing the unnecessary nesting and using early returns/guards for error handling. It’ll make the flow flatter and easier to read instead of chaining multiple `if/else` blocks.

internal/integration/whop/invoice_sync.go
```diff
@@ -0,0 +1,306 @@
...
+	inv, err := s.invoiceRepo.Get(ctx, invoiceID)
+	if err != nil {
+		return err
+	}
+	if inv.Metadata == nil {
+		inv.Metadata = make(types.Metadata)
+	}
+	inv.Metadata["whop_checkout_url"] = purchaseURL
+	return s.invoiceRepo.Update(ctx, inv)
+}
```
omkar273: We can remove the unnecessary helper method since it’s only used in a single place and replace it with an inline implementation to keep the flow simpler and easier to follow.

internal/temporal/client/client.go
```diff
@@ -81,14 +88,15 @@ func (c *temporalClient) Start(ctx context.Context) error {
...
-	// Check health to ensure connection is working
-	if _, err := c.client.CheckHealth(ctx, &client.CheckHealthRequest{}); err != nil {
-		c.logger.Error("Failed to check client health during start", "error", err)
-		return err
+	healthCtx, cancel := context.WithTimeout(ctx, 2*time.Second)
+	defer cancel()
+	if _, err := c.client.CheckHealth(healthCtx, &client.CheckHealthRequest{}); err != nil {
+		c.logger.Warn("Temporal health check failed during start; continuing — workflows will retry on their own",
+			"error", err)
 	}
```
omkar273, line 88: Why are we changing this ??

internal/temporal/client/client.go
```diff
@@ -72,7 +75,11 @@ func NewTemporalClient(options *models.ClientOptions, logger *logger.Logger) (Te
 	}, nil
 }
 
-// Start implements TemporalClient
+// Start implements TemporalClient.
+// The health check is best-effort: a transient cold-start failure should not block
+// Fx dependency injection. The underlying gRPC connection (via NewLazyClient) is
+// dialled on first actual RPC and has its own retries, so real connectivity
+// problems will surface there.
```
omkar273: revert all changes to related to temporal client

internal/integration/whop/client.go
subratsahilgupta: No point!

internal/integration/whop/dto.go
subratsahilgupta: good point

internal/integration/whop/dto.go
subratsahilgupta: decimal.Decimal produces quoted string, whop rejects that. So float64 is the right type for the outbound DTO

PR #1925, opened 2026-06-05
ci: push to GHCR on every deploy, not just tag releases

agrim123:
## Summary

- **Always push to GHCR** on every `main`/`develop` push (previously only happened on `v*` tag releases). Every build now gets `ghcr.io/<repo>:<sha>` tagged in GHCR.
- **Isolated GHCR failures** — split the single combined build+push step into two: ECR push (blocking, deploy depends on it) and GHCR push (`continue-on-error: true`). A GHCR outage or auth failure will no longer break ECS deploys.
- **Auth consistency** — switched GHCR login from `GITHUB_TOKEN` to `GHCR_PUBLISH_TOKEN` + `GHCR_USERNAME` var, matching `publish-app-image.yml`.

## Behaviour unchanged

- ECR push and all ECS deploy jobs work exactly as before.
- Semver tags (`ghcr.io/<repo>:v1.2.3`) and `:latest` on GHCR are still release-only.
- Build platforms (`linux/arm64` for branch pushes, multi-arch for releases) unchanged.

## Test plan

- [x] Push to `develop` — confirm ECR image pushed and ECS deploys succeed; confirm GHCR SHA-tagged image appears
- [x] Simulate GHCR auth failure (e.g. bad token) — confirm ECS deploy still succeeds and GHCR step shows as a warning, not a failure
- [x] Push a `v*` tag — confirm ECR + GHCR both get semver and SHA tags, `:latest` updated on GHCR

## Summary by CodeRabbit

* **Chores**
  * Optimized deployment workflow to separate publishing for different image registries, improving reliability and error tolerance during container pushes.
  * Improved image tagging and publishing steps while preserving existing build outputs and signatures to avoid disrupting downstream integrations.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -62,6 +62,7 @@ jobs:
         uses: aws-actions/amazon-ecr-login@v2
 
       - name: Login to GitHub Container Registry
+        continue-on-error: true
         uses: docker/login-action@v3
         with:
           registry: ghcr.io
@@ -100,7 +101,7 @@ jobs:
 
       - name: Build & push image to ECR
         id: build-ecr
-        uses: docker/build-push-action@v4
+        uses: docker/build-push-action@v7
         with:
           context: .
           file: Dockerfile
@@ -110,7 +111,7 @@ jobs:
           provenance: false
 
       - name: Push image to GHCR
-        uses: docker/build-push-action@v4
+        uses: docker/build-push-action@v7
         continue-on-error: true
         with:
           context: .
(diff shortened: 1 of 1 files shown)


agrim123:
Fixed in the latest commit:
- Added `continue-on-error: true` to the GHCR login step so auth failures don't block ECR push or ECS deploys
- Bumped `docker/build-push-action` from `v4` to `v7` (latest stable) on both ECR and GHCR push steps

nkmishra1997 (approved):

PR #1933, opened 2026-06-05
fix(kafka): support GCP Managed Service for Apache Kafka (GMK) over SASL/OAUTHBEARER

agrim123:
## Problem

Connecting the app to **GCP Managed Service for Apache Kafka (GMK)** over `SASL/OAUTHBEARER` failed in two ways:

1. The broker rejected every connection with:
   ```
   kafka server: SASL Authentication failed: Authentication failed during
   authentication due to invalid credentials with SASL mechanism OAUTHBEARER
   ```
2. Once the events client was fixed, the wallet-alert/onboarding client crashed at startup with:
   ```
   kafka: invalid configuration (An AccessTokenProvider instance must be
   provided to Net.SASL.TokenProvider)
   ```

## Root causes

**1. Wrong token format.** GMK does **not** accept a bare OAuth2 access token over OAUTHBEARER. It expects a Google-specific token — three base64url segments joined with `.`:
- header `{"typ":"JWT","alg":"GOOG_OAUTH2_TOKEN"}`
- payload `{"exp","iss":"Google","iat","sub":<principal email>}`
- the raw access token

This matches Google's reference implementation:
https://github.com/googleapis/managedkafka/blob/main/kafka-auth-local-server/kafka_gcp_credentials_server.py

`internal/kafka/oauth.go` previously returned the plain access token.

**2. Two Kafka clients.** The app has two sarama-based clients — `internal/kafka` (events) and `internal/pubsub/kafka` (wallet alerts, onboarding). The second set `Net.SASL.Mechanism = OAUTHBEARER` but never attached a `TokenProvider`.

## Changes

- `internal/kafka/oauth.go`: emit the GMK `GOOG_OAUTH2_TOKEN` JWT token. Principal (`sub`) resolved from the metadata server (impersonated GSA under Workload Identity), cached, overridable via `GOOGLE_MANAGED_KAFKA_AUTH_PRINCIPAL`. Export `NewGCPTokenProvider`.
- `internal/pubsub/kafka/base.go`: wire the same shared token provider for the OAUTHBEARER case.
- `internal/ee/.../helm/.../_helpers.tpl`: only inject the `kafka-sasl-password` `secretKeyRef` for password-based mechanisms — OAUTHBEARER has no password, and `secret.yaml` only renders that key when `saslPassword` is set, so pods failed with `CreateContainerConfigError`.
- `internal/kafka/oauth_test.go`: verifies the token wire format byte-for-byte against the reference.

## Verification

- Unit test passes (token format).
- Validated live on a GCP staging GKE cluster against a real GMK cluster: all app pods healthy, `INVALID_TOPIC`/auth errors gone, broker shows the consumer group registered (both client paths authenticate and consume).

## Summary by CodeRabbit

* **New Features**
  * Added Google Managed Kafka OAuth/OAUTHBEARER support with GMK-compatible token format.
  * Kafka client now selects SASL behavior by mechanism (OAUTHBEARER, SCRAM, PLAIN).
  * Helm charts/values now include explicit enabled flags for event consumers and feature-usage tracking.

* **Bug Fixes / Behavior**
  * Kafka SASL password env var is omitted for token-based OAUTHBEARER flows.

* **Tests**
  * Added tests verifying the GMK OAuth token wire format.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -86,7 +86,7 @@ require (
 require (
 	ariga.io/atlas v0.19.1-0.20240203083654-5948b60a8e43 // indirect
 	cel.dev/expr v0.25.1 // indirect
-	cloud.google.com/go/compute/metadata v0.9.0 // indirect
+	cloud.google.com/go/compute/metadata v0.9.0
 	github.com/ClickHouse/ch-go v0.66.0 // indirect
 	github.com/KyleBanks/depth v1.2.1 // indirect
 	github.com/agext/levenshtein v1.2.1 // indirect
--- a/internal/ee/infrastructure/helm/flexprice/templates/_helpers.tpl
+++ b/internal/ee/infrastructure/helm/flexprice/templates/_helpers.tpl
@@ -322,11 +322,21 @@ All service addresses are resolved via named templates above so this block stays
   value: {{ .Values.kafkaConfig.saslMechanism | quote }}
 - name: FLEXPRICE_KAFKA_SASL_USER
   value: {{ .Values.kafkaConfig.saslUser | quote }}
+{{- /*
+  OAUTHBEARER (e.g. GCP Managed Kafka) carries no static password — the token
+  comes from a sarama AccessTokenProvider (Application Default Credentials under
+  Workload Identity). Only wire the password secretKeyRef for password-based
+  mechanisms (PLAIN / SCRAM); otherwise the env references a kafka-sasl-password
+  key that secret.yaml never renders (it gates on .saslPassword), which fails
+  the pod with CreateContainerConfigError.
+*/}}
+{{- if ne .Values.kafkaConfig.saslMechanism "OAUTHBEARER" }}
 - name: FLEXPRICE_KAFKA_SASL_PASSWORD
   valueFrom:
     secretKeyRef:
       name: {{ include "flexprice.secretName" . }}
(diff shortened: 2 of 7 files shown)


agrim123:
Thanks @coderabbitai — good catch on the structural inconsistency. Fixed in 1d0838c7: `Net.SASL.User`/`Password` are now set only inside the SCRAM and `default` (PLAIN) cases, matching `internal/kafka/base.go`. The OAUTHBEARER path no longer touches them (it doesn't use them). No behavior change; builds and vets clean.

agrim123:
Added a related Helm fix in 96ce6654: the chart's configmap rendered the `event_processing` / `event_processing_lazy` / `event_post_processing` / `feature_usage_tracking(_lazy)` blocks **without an `enabled` field**. Those Go structs default `Enabled` to `true`, but viper ignores Go `default` tags on Unmarshal — so `Enabled` resolved to `false` and the **events→ClickHouse consumers never subscribed**. Same root cause/shape as the onboarding_events fix (#1914).

Verified end-to-end on GCP staging after this fix: `POST /v1/events` → 202 → published to GMK `events` topic over OAUTHBEARER → consumer (now subscribed to `events`) → rows land in ClickHouse and are returned by `GET /v1/events`.

nkmishra1997 (approved):

PR #1945, opened 2026-06-09
feat(schedule): pick db lookup reference window as schedule job time

Gursimar10:
## Summary by CodeRabbit

* **Refactor**
  * Subscription renewal alert flow now uses an explicit reference time across services and workflows for better timing control and testability.
  * The renewal-due window calculation was adjusted (narrower pre-renewal window), which may slightly change when renewal alerts are emitted.

### DIFF
--- a/internal/domain/subscription/repository.go
+++ b/internal/domain/subscription/repository.go
@@ -2,6 +2,7 @@ package subscription
 
 import (
 	"context"
+	"time"
 
 	"github.com/flexprice/flexprice/internal/types"
 )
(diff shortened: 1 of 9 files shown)


internal/domain/subscription/repository.go
```diff
@@ -27,7 +28,7 @@ type Repository interface {
 	GetWithPauses(ctx context.Context, id string) (*Subscription, []*SubscriptionPause, error)
 
 	// Renewal due alert methods
-	ListSubscriptionsDueForRenewal(ctx context.Context) ([]*Subscription, error)
+	ListSubscriptionsDueForRenewal(ctx context.Context, now time.Time) ([]*Subscription, error)
```
omkar273: We can improve the name of the variable rather than now

internal/interfaces/service.go
```diff
@@ -117,7 +117,7 @@ type SubscriptionService interface {
 	// Auto-cancellation methods
 	ProcessAutoCancellationSubscriptions(ctx context.Context) error
 	// Renewal due alert methods
-	ProcessSubscriptionRenewalDueAlert(ctx context.Context) error
+	ProcessSubscriptionRenewalDueAlert(ctx context.Context, now time.Time) error
```
omkar273: We can improve the name of the variable rather than now

omkar273 (approved):

PR #1962, opened 2026-06-10
feat(observability): OTel-native error & exception reporting to SigNoz

nkmishra1997:
## Summary

Migrates error & exception capture from **Sentry** to **OTel-native `exception` span events**, so errors surface in **SigNoz's Exceptions tab** (per https://signoz.io/docs/userguide/exceptions/). Sentry is **gated off** but kept in-tree for rollback — deleting the deps/config is a deliberate follow-up.

Traces and logs were already OTLP-native to SigNoz; the only remaining Sentry-only path was explicit error capture. This closes that gap.

## What changed

- **New `internal/spanerr` leaf package** — emits `exception` span events (`exception.type` / `exception.message` / `exception.stacktrace`, the attributes SigNoz reads) with best-effort per-span dedup and an always-captured stacktrace (which `RecordError` omits by default).
- **`tracing.CaptureException(ctx, err)`** — now takes `ctx`; records onto the active span, or **synthesizes** a short-lived `error.capture` span when there's none (background goroutines, consumers). `AddBreadcrumb(ctx, …)` → span events.
- **Logger auto-capture** — `log.Error(ctx, …)` records an exception event when ctx carries a recording span. Gated by `otel.traces.capture_exceptions` (default `true`).
- **Panics** — new `OtelRecoveryMiddleware` records the panic (with stacktrace) then re-panics to gin's outer recovery. Replaces `sentrygin`.
- **Dedup scopes** seeded in the HTTP request (`SpanEnrichmentMiddleware`), `StartTransaction`, and the Temporal activity interceptor so a logged-and-captured error yields one event.
- ~10 capture sites threaded `ctx`; Temporal activity capture dropped (its `SetStatusError` already records, now with a stacktrace).
- `sentry.enabled` default flipped to `false`.

## Not in this PR (follow-up)

Deleting `getsentry/sentry-go` deps, `SentryConfig`, and `initSentry`/Sentry-`Flush` — once SigNoz exceptions are validated in prod.

## Testing

- `go build ./internal/... ./cmd/...` ✅ · `go vet ./internal/...` ✅
- New `internal/spanerr` unit tests (event emission, dedup, panic, no-span no-op) ✅
- Local end-to-end validation against an OTLP collector (see PR comments).

## Rollout note

Keep traces `sample_rate: 1.0` so error-bearing traces aren't sampled away.

## Summary by CodeRabbit

* **Refactor**
  * Migrate error/exception handling to OpenTelemetry-native capture; panics now emit OTel exception events and per-request dedup scopes reduce duplicates.
* **Configuration**
  * OTel exception capture enabled by default; legacy error backend disabled by default and retained only for rollback.
* **Bug Fixes**
  * Improved context propagation so exception events attach to active requests and background flows; reduced duplicate recordings.
* **Tests**
  * Added unit tests for exception recording, panic capture, and deduplication.

### DIFF
--- a/internal/api/router.go
+++ b/internal/api/router.go
@@ -93,12 +93,16 @@ func NewRouter(
 		middleware.LoggingMiddleware(logger), // Use our standard logger for HTTP logging
 		middleware.CORSMiddleware,
 	)
-	// Tracing middleware fans out into otelgin (SigNoz / OTLP) and sentrygin
-	// (panic recovery + Sentry scope binding). Each handler is added separately
-	// because gin's Use signature is variadic and the slice may be empty.
+	// Tracing middleware creates the otelgin span per request (SigNoz / OTLP).
+	// Each handler is added separately because gin's Use signature is variadic
+	// and the slice may be empty.
 	for _, h := range middleware.TracingMiddleware(cfg) {
 		router.Use(h)
 	}
+	// OtelRecoveryMiddleware runs inside the otelgin span so it can record a
+	// panic as an exception span event before re-panicking to gin's outer
+	// recovery (registered above) for the 500 response + log.
+	router.Use(middleware.OtelRecoveryMiddleware())
 	// SpanEnrichmentMiddleware runs after otelgin (span created) and before handlers.
 	// Post-phase executes before otelgin's post-phase (LIFO), so it can set span
 	// status / record errors before the span is ended and exported.
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -126,22 +126,22 @@ type SupabaseConfig struct {
 }
 
 type KafkaConfig struct {
-	Brokers                []string             `mapstructure:"brokers" validate:"required"`
-	ConsumerGroup          string               `mapstructure:"consumer_group" validate:"required"`
-	Topic                  string               `mapstructure:"topic" validate:"required"`
-	TopicLazy              string               `mapstructure:"topic_lazy" validate:"required"`
-	TopicDLQ               string               `mapstructure:"topic_dlq" default:""`
-	TLS                    bool                 `mapstructure:"tls"` // set to true if using 9094 port else can set to false
-	UseSASL                bool                 `mapstructure:"use_sasl"`
-	SASLMechanism          sarama.SASLMechanism `mapstructure:"sasl_mechanism"`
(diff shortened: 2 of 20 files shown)


nkmishra1997:
## Local validation (real OTLP collector)

Ran the full app locally (`FLEXPRICE_DEPLOYMENT_MODE=local`) with traces pointed at a standalone `otel/opentelemetry-collector` (debug exporter) and `capture_exceptions=true`, then triggered real errors. Both capture paths verified end-to-end over OTLP/gRPC:

**1. On-span auto-capture** (server-side 500 inside an HTTP request span)
```
Span  http.route=/v1/customers/:id  http.response.status_code=500
SpanEvent #0
  -> Name: exception
  -> exception.type: *markers.withMark
  -> exception.message: ent: tenant not found
  -> exception.stacktrace: ... spanerr.RecordException -> logger.logErrorToSpan
                               -> logger.Error -> tenant_access.go:27
```

**2. Synthesized `error.capture` span** (spanless capture — consumer hit a poison message)
```
Span  Name=error.capture  Kind=Internal  Status code=Error
SpanEvent #0
  -> Name: exception
  -> exception.type: *json.SyntaxError
  -> exception.message: unexpected end of JSON input
  -> exception.stacktrace: ... tracing.Service.CaptureException (tracing.go:362)
                               -> eventConsumptionService.processMessage (event_consumption.go:209)
```

Also confirmed: span status flips to `Error`, per-request dedup yields one event per request (no otelgin duplication), and startup logs show `Sentry is disabled` + `OTel tracing initialized`.

### Bug caught & fixed during validation
With traces enabled but no exception events appearing, root cause was that `FLEXPRICE_OTEL_TRACES_CAPTURE_EXCEPTIONS` never bound — viper's `AutomaticEnv` skips nested keys with underscores (the codebase already `BindEnv`s every other `otel.traces.*` key for exactly this reason). Under env-driven prod config the feature would have been silently off. Fixed in `71b5b4c3` with an explicit `BindEnv` + `SetDefault(true)`.

### Minor follow-up note (not blocking)
For errors logged as `"error", err` (raw), `exception.type` resolves to the wrapper type `*markers.withMark`. Logging via `logger.Err(err)` (which adds `error_type`) yields a more specific type for SigNoz grouping; `exception.message` already differentiates them, so default grouping still works.

PR #1968, opened 2026-06-11
feat(kafka): kafka-migrate reconciler — image-baked topic reconciliation

agrim123:
## What

Adds an image-baked `/app/kafka-migrate` entrypoint that reconciles Kafka topics from a committed `topics.yaml` — forward-only and non-destructive (create + grow-partitions, never shrink, never delete). Reuses the existing GMK OAUTHBEARER admin client path (`GetSaramaConfig` + `oauth.go` token provider).

## How

- **`internal/kafka/topicspec`** — parses an env-agnostic `topics.yaml` (topic names + default sizing) and resolves it against per-env overrides injected by Helm via env-vars (`FLEXPRICE_KAFKA_TOPIC_<TOPIC>_PARTITIONS|REPLICATIONFACTOR|RETENTIONMS`). Repo file stays env-agnostic.
- **`internal/kafka/reconcile`** — Kafka-client-agnostic diff/apply algorithm behind a small `Admin` interface. `SaramaAdmin` is one adapter over `sarama.ClusterAdmin`; others can be added without touching the algorithm. Create missing, grow partitions when fewer; warn-only (never act) on shrink and RF mismatch.
- **`cmd/kafka-migrate/main.go`** — wiring only: `config.NewConfig()` → `GetSaramaConfig(cfg)` → `ClusterAdmin` → load+resolve `topics.yaml` → reconcile. Supports `--dry-run` (logs WOULD CREATE / WOULD GROW / WARN). Non-zero exit on reconcile failure so a hook Job blocks the rollout.
- **`Dockerfile`** — builds `kafka-migrate` and copies it + `topics.yaml` to `/app`, matching `/app/migrate`.

## Tests

- `topicspec`: 6 table tests (parse, defaults, per-topic + env-var override precedence, zero-partition rejection). PASS.
- `reconcile`: 7 tests against a fake `Admin` (create / grow / skip-shrink / unmanaged-untouched / RF-mismatch-warn / grow-error propagation) + sarama mapping. PASS.
- Image builds; `/app/kafka-migrate` + `/app/topics.yaml` verified present.

Also synced an orphaned `creasty/defaults` vendor entry (`go mod tidy && go mod vendor`) that was breaking clean `docker build`.

## Out of scope (follow-on)

CH golang-migrate swap; infra-repo Helm hook Job wiring + Terraform topic removal; IAM/Workload-Identity admin grant. Staging GMK dry-run pending cluster access.

## Notes

- `topics.yaml` currently lists 17 topics derived from `KafkaConfig` `default:` tags — confirm completeness against all producer/consumer topics before a prod hook fires.
- RF is warn-only by design (managed-Kafka RF fidelity).

## Summary by CodeRabbit

* **New Features**
  * Added a Kafka topic reconciliation CLI with a --dry-run preview mode and --allow-defaults startup flag
  * Support for environment-variable overrides (FLEXPRICE_KAFKA_TOPIC_*) to customize partitions, replication factor, and retention
  * Startup validation that requires topic configuration; emits warnings for replication-factor mismatches and skipped partition shrinks

* **Chores**
  * Updated runtime image to include the reconciliation binary
  * Dependency classification updated for YAML handling


agrim123:
@coderabbitai addressed in 5f2c5f4c:

- **🟠 main.go:63 — dry-run duplicates reconcile logic / misses RF mismatch.** Fixed properly. `Reconcile` is now split into `Plan(admin, desired) ([]Action, error)` (pure decisions, RF mismatch included) and `Apply(admin, plan) (Result, error)`. `main.go` builds the plan once; dry-run logs the *same* plan a live apply would execute — so it can no longer drift, and dry-run now emits the RF-mismatch warning too. `Reconcile` is kept as `Plan`+`Apply` for the existing API.
- **🟡 spec.go:64/52/89 — unused `env` param + no post-override validation.** `Resolve` now returns an error, re-validates the *resolved* values (after defaults + per-topic + env-var overrides) and rejects `partitions < 1` / `replicationFactor < 1` — so an env-injected `PARTITIONS=0` no longer slips past `Parse`. `env` is now used in the error context. `Parse` also rejects a per-topic `replicationFactor < 1`.
- **🟠 reconcile_test.go:89 — table-driven tests.** The six `TestReconcile_*` functions are collapsed into one table-driven `TestReconcile` (per-case `checkAdmin` closure), plus a new `TestPlan_EmitsRFMismatchAndGrow` verifying a topic that is both under-partitioned and RF-mismatched yields both actions.
- **🟡 spec_test.go:40/47 — `_`-ignored Parse errors.** Replaced with `require.NoError`; added `TestResolve_RejectsZeroPartitionOverride` and `TestParse_RejectsZeroReplicationFactor`.

The golangci-lint `api/custom/go` typecheck errors flagged in the inline tool output are pre-existing and unrelated to this PR. All new tests pass; `go vet` clean.

agrim123:
## Update: config-derived topics (replaces topics.yaml)

After verifying against **both prod MSK clusters** (us-west-2 `prod-flexprice-msk-us`, ap-south-1 `prod-flepxrice-v2`), the hand-maintained `topics.yaml` model was wrong: the clusters are **shared across envs for cost**, and topic names are namespaced per-env via env-vars (`prod_system_events`, `balance_alert`, `staging_*`). A static base-name yaml would create phantom unprefixed topics on prod.

**New approach — derive desired topics from the resolved app config:**
- `topicspec.FromConfig(cfg)` reflection-harvests every field tagged `topic`/`topic_lazy`/`topic_dlq`/`topic_backfill`/`output_topic` across the whole `*config.Configuration`, skips empties, dedups.
- Names come from `config.NewConfig()` where **env-vars already win** (viper) → correct per-env names automatically, on every cluster, with zero per-env yaml.
- Sizing: `FLEXPRICE_KAFKA_TOPIC_<NAME>_*` env override (highest) → default 6 partitions / rf3 / 7d.
- **Startup guard**: refuses to run (exit 1, no mutation) if no `FLEXPRICE_*_TOPIC` env is set — prevents falling back to unprefixed struct defaults and creating phantom topics on a shared cluster. `--allow-defaults` for local/dev.
- `topics.yaml` deleted (code + image). Orphan topics still left untouched (forward-only).

**Operational requirement:** the kafka-migrate Job must inherit the app deployment's env block (same topic env-vars), as the staging dry-run did. Design doc + plan capture this as a hard precondition.

**Known drift surfaced:** prod-us env sets `FLEXPRICE_RAW_EVENTS_PROCESSING_OUTPUT_TOPIC=prod_events_v4` but the live cluster has `prod_events_v5` — resolve before any prod apply; always `--dry-run` first.

Verify: all 3 kafka packages pass, `go vet` clean, image builds, guard fires on bare env (exit 1, no broker connect).

agrim123:
@coderabbitai addressed in 85f2cc5c: `fromConfigStruct` now validates resolved sizing after override application (rejects `partitions < 1` / `replicationFactor < 1`) and `FromConfig` returns an error, which `main.go` handles with `log.Fatalf` — so an env-injected `FLEXPRICE_KAFKA_TOPIC_<NAME>_PARTITIONS=0` fails fast with a precise message instead of surfacing later during reconcile/apply. Added `TestFromConfigStruct_RejectsZeroPartitionOverride` + `_RejectsZeroRFOverride`.

agrim123:
## Important correction: create-only partitions (9a4d0e16)

Verified live partition counts across clusters — they are **hand-tuned per env and live only on the cluster**, NOT in config/env (no `_PARTITIONS` env-var anywhere):

| topic | staging | prod-us |
|---|---|---|
| events | 6 | **100** |
| events_lazy | 6 | **40** |
| events_backfill | 6 | **20** |
| system_events | 3 | 6 |

A default-and-grow model would have been **destructive** — a default of 6 against any topic whose live count is below it would GROW partitions (irreversible; reshuffles key→partition mapping).

**Fix:** reconciler is now **create-only for partitions**:
- Existing topic → always `Unchanged` (never compare/grow/shrink partitions; live count is the truth). RF mismatch still warns.
- Missing topic → created with env-override size if set, else default 6/rf3/7d. New topics start small; ops grows them (grow is non-destructive, over-provisioning isn't).
- Removed the `Grow`/`SkipShrink` mutation paths; only `ActionCreate` mutates.

The post-override validation from the previous commit still guards `create` sizing. Tests updated to assert existing topics are never grown/shrunk.

agrim123:
## Redesign v3: topics.yaml base + JSON env override (replaces config-reflection)

Reverted the config-reflection approach. New model is simpler and explicit:

**Source resolution:**
- Baked **`topics.yaml`** (repo, name-keyed) is the base/reference = the 9 staging GMK topics.
- **`FLEXPRICE_KAFKA_TOPICS`** — a JSON spec (same schema) delivered per-env via ECS env-var / k8s ConfigMap. When set it **fully replaces** the baked file (no merge). When unset, the baked yaml is used.

**Schema (yaml + JSON identical), keyed by topic name:**
```yaml
defaults: { replicationFactor: 3, retentionMs: 604800000 }
topics:
  events: { partitions: 6 }
  events_dlq: { partitions: 3 }
```

**Reconcile — create + grow:** create missing; grow when spec > live; skip+warn when spec < live (never shrink); RF warn-only; never delete. Sizing is now operator-authored (yaml/JSON), so growth is intentional — not a guessed default. (To run a pure no-op on an existing cluster, set the override's counts to match live.)

**Removed:** the v2 reflection harvester (`FromConfig`), the per-name `FLEXPRICE_KAFKA_TOPIC_*_PARTITIONS` env mechanism, and the startup guard (no struct-default fallback to guard now).

**Verified:**
- override env set → `topics=1`, `prod_events_v5@100` (JSON drives it, file ignored)
- no override → `topics=9` baked staging yaml (events@6, dlq@3, …)
- all 3 kafka packages pass; `go vet` clean; image builds with `/app/kafka-migrate` + `/app/topics.yaml`.

Design + plan: castle `Flexprice/specs|plans/2026-06-11-*`.

agrim123:
## Per-env `FLEXPRICE_KAFKA_TOPICS` values (AWS)

> **Nothing breaks if the migration job is never run.** `/app/kafka-migrate` is baked into the image but is NOT wired to any startup/hook/entrypoint — the app deploys and runs exactly as before and never touches it. The reconciler only does anything when you explicitly run it as a one-off ECS task / k8s job. So these env-vars are inert until that job runs, and even then `--dry-run` shows the plan before any apply.

Each value below = that env's app topic names (harvested from its ECS task-def `*_TOPIC` env-vars) sized to the **live cluster partition counts**, so a real run is a no-op (no grow/shrink) — it would only create a topic that's genuinely missing. Set as the `FLEXPRICE_KAFKA_TOPICS` env on the migrate task for each env.

### US-WEST-2 PROD — `prod-flexprice-msk-us`
```
FLEXPRICE_KAFKA_TOPICS={"defaults":{"replicationFactor":3,"retentionMs":604800000},"topics":{"balance_alert":{"partitions":20},"events":{"partitions":100},"events_backfill":{"partitions":20},"events_lazy":{"partitions":40},"prod_benchmarking":{"partitions":5},"prod_events_v4":{"partitions":6},"prod_raw_events":{"partitions":20},"prod_system_events":{"partitions":6}}}
```
⚠️ `prod_events_v4` is in the app env-var but the live cluster has `prod_events_v5` (no v4). As-is a run would CREATE `prod_events_v4@6`. Resolve the app env-var (v4→v5) or drop it from this override before applying.

### AP-SOUTH PROD — `prod-flepxrice-v2` (shared prod+staging cluster)
```
FLEXPRICE_KAFKA_TOPICS={"defaults":{"replicationFactor":3,"retentionMs":604800000},"topics":{"balance_alert":{"partitions":20},"events":{"partitions":20},"events_backfill":{"partitions":6},"events_lazy":{"partitions":20},"prod_benchmarking":{"partitions":5},"prod_events_dlq":{"partitions":10},"prod_system_events":{"partitions":6}}}
```
All 7 match live → pure no-op.

### AP-SOUTH STAGING — `prod-flepxrice-v2` (same cluster, staging_ prefix)
```
FLEXPRICE_KAFKA_TOPICS={"defaults":{"replicationFactor":3,"retentionMs":604800000},"topics":{"staging_benchmarking":{"partitions":2},"staging_bulk_events":{"partitions":2},"staging_events":{"partitions":3},"staging_events_backfill":{"partitions":3},"staging_events_lazy":{"partitions":3},"staging_events_v2":{"partitions":3},"staging_onboarding_events":{"partitions":2},"staging_system_events":{"partitions":2}}}
```
All 8 match live → pure no-op.

### Notes
- The ap-south v2 cluster hosts both prod and staging via topic-name prefixes, so prod and staging each get their own override on their respective task-defs.
- Partition counts are per-cluster tuned (us-west2 `events`=100 vs ap-south `events`=20).
- GCP staging (GMK) is separate (9 topics, already validated) — not in this AWS list.
- Names were harvested from each env's api/consumer task-def; if a worker/other deploy uses a topic the api doesn't, it isn't here — union across all task-defs per env before relying on it for completeness.

internal/config/config.yaml
```diff
@@ -33,6 +33,34 @@ kafka:
...
   route_tenants_on_lazy_mode: []
+  # topics_defaults/topics below describe the full desired topic set for
+  # cmd/kafka-migrate only (server/consumer/worker ignore them). Base/dev
+  # names, unprefixed — mirrors the 9 staging GMK topics. A shared prod
+  # cluster MUST set FLEXPRICE_KAFKA_TOPICS (JSON, same shape) which FULLY
+  # REPLACES this block (no merge) with prod_* prefixed names + tuned sizing.
+  # See internal/kafka/topicspec and topics.example.json.
+  topics_defaults:
+    replication_factor: 3
+    retention_ms: 604800000 # 7d
```
ojasagarwal-flex, line 44: we should make this as well topic wise?

agrim123: it is already, these are just defaults

ojasagarwal-flex (approved):

PR #1984, opened 2026-06-12
feat(integration): delink api to remove provider entity mapping

Gursimar10:
## Summary by CodeRabbit

* **New Features**
  * Added ability to delink (archive) integration mappings via a new authenticated API endpoint.
  * Returns a confirmation response including the number of mappings archived.

* **Bug Fixes**
  * Ensures delinking returns a clear “not found” error when no matching mappings exist.

* **Tests**
  * Added unit tests covering successful archiving/removal and the not-found scenario.

### DIFF
--- a/internal/api/dto/integration.go
+++ b/internal/api/dto/integration.go
@@ -85,8 +85,8 @@ func (r *LinkIntegrationMappingRequest) Validate() error {
 
 type DelinkIntegrationMappingRequest struct {
 	EntityType   types.IntegrationEntityType `json:"entity_type" validate:"required"`
-	EntityID     string                      `json:"entity_id" validate:"required,max=255"`
-	ProviderType string                      `json:"provider_type" validate:"required,max=50"`
+	EntityID     string                      `json:"entity_id" validate:"required"`
+	ProviderType string                      `json:"provider_type" validate:"required"`
 }
 
 type DelinkIntegrationMappingResponse struct {
--- a/internal/api/dto/integration.go
+++ b/internal/api/dto/integration.go
@@ -83,6 +83,30 @@ func (r *LinkIntegrationMappingRequest) Validate() error {
 	return nil
 }
 
+type DelinkIntegrationMappingRequest struct {
+	EntityType   types.IntegrationEntityType `json:"entity_type" validate:"required"`
+	EntityID     string                      `json:"entity_id" validate:"required,max=255"`
+	ProviderType string                      `json:"provider_type" validate:"required,max=50"`
+}
+
+type DelinkIntegrationMappingResponse struct {
+	Success  bool `json:"success"`
+	Archived int  `json:"archived"`
+}
+
+func (r *DelinkIntegrationMappingRequest) Validate() error {
+	if err := validator.ValidateRequest(r); err != nil {
+		return err
+	}
+	if err := r.EntityType.Validate(); err != nil {
+		return err
+	}
+	if err := types.SecretProvider(r.ProviderType).Validate(); err != nil {
+		return err
+	}
+	return nil
+}
+
 type IntegrationConfigEntry struct {
 	Provider      types.SecretProvider `json:"provider"`
 	BaseConfig    *types.SyncConfig    `json:"base_config"`
--- a/internal/api/router.go
+++ b/internal/api/router.go
@@ -530,6 +530,7 @@ func NewRouter(
 		integrations := v1Private.Group("/integrations")
 		{
 			integrations.POST("/link", handlers.Integration.Link)
+			integrations.POST("/delink", handlers.Integration.Delink)
 			integrations.POST("/sync", handlers.Integration.Sync)
 			integrations.GET("/mappings", handlers.Integration.GetMappings)
 			integrations.GET("/config", handlers.Integration.GetConfig)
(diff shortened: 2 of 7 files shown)


internal/service/entityintegrationmapping.go
```diff
@@ -268,6 +268,47 @@ func (s *entityIntegrationMappingService) LinkIntegrationMapping(ctx context.Con
...
+			WithReportableDetails(map[string]any{
+				"entity_id":     req.EntityID,
+				"entity_type":   req.EntityType,
+				"provider_type": req.ProviderType,
+			}).
+			Mark(ierr.ErrNotFound)
+	}
+
+	for _, mapping := range mappings {
+		if err := s.EntityIntegrationMappingRepo.Delete(ctx, mapping); err != nil {
```
Gursimar10: marks the row archived

internal/api/dto/integration.go
```diff
@@ -83,6 +83,30 @@ func (r *LinkIntegrationMappingRequest) Validate() error {
 	return nil
 }
 
+type DelinkIntegrationMappingRequest struct {
+	EntityType   types.IntegrationEntityType `json:"entity_type" validate:"required"`
+	EntityID     string                      `json:"entity_id" validate:"required,max=255"`
+	ProviderType string                      `json:"provider_type" validate:"required,max=50"`
```
omkar273: We should remove the max length validations

internal/api/dto/integration.go
```diff
@@ -83,6 +83,30 @@ func (r *LinkIntegrationMappingRequest) Validate() error {
 	return nil
 }
 
+type DelinkIntegrationMappingRequest struct {
+	EntityType   types.IntegrationEntityType `json:"entity_type" validate:"required"`
```
omkar273, line 87: For now, these fields are good enough we can extend this to also accept `provider_entity_id`

internal/api/v1/integration.go
```diff
@@ -86,6 +86,37 @@ func (h *IntegrationHandler) Link(c *gin.Context) {
...
+func (h *IntegrationHandler) Delink(c *gin.Context) {
+	var req dto.DelinkIntegrationMappingRequest
+	if err := c.ShouldBindJSON(&req); err != nil {
+		c.Error(ierr.WithError(err).WithHint("Invalid request body").Mark(ierr.ErrValidation))
+		return
+	}
+	if err := req.Validate(); err != nil {
+		c.Error(err)
+		return
+	}
```
omkar273: We should remove this from handler layer and move this to service layer

internal/integration/paddle/sync_service_test.go
```diff
@@ -117,6 +117,10 @@ func (s *inMemoryMappingService) LinkIntegrationMapping(ctx context.Context, req
 	return nil, nil
 }
 
+func (s *inMemoryMappingService) DelinkIntegrationMapping(ctx context.Context, req apidto.DelinkIntegrationMappingRequest) (*apidto.DelinkIntegrationMappingResponse, error) {
+	return nil, nil
+}
```
omkar273, line 153: We should add implementation for this, i.e. seed some `entity integration mapping` and test them

internal/service/entityintegrationmapping.go
```diff
@@ -268,6 +268,47 @@ func (s *entityIntegrationMappingService) LinkIntegrationMapping(ctx context.Con
...
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
omkar273: Valid comment

internal/api/router.go
```diff
@@ -532,6 +530,7 @@ func NewRouter(
 		integrations := v1Private.Group("/integrations")
 		{
 			integrations.POST("/link", handlers.Integration.Link)
+			integrations.POST("/delink", handlers.Integration.Delink)
```
omkar273: Lets change to `DELETE /integrations/link`

internal/api/dto/integration.go
Gursimar10: Removed the `max` length validations from `DelinkIntegrationMappingRequest` (entity_id and provider_type).

internal/api/v1/integration.go
Gursimar10: Removed `req.Validate()` from the handler; validation now lives in the service layer (`DelinkIntegrationMapping` already calls `req.Validate()` first).

internal/integration/paddle/sync_service_test.go
Gursimar10: Implemented the test double's `DelinkIntegrationMapping` to mirror the real service (list by entity+provider, NotFound when none, delete each) and added TestDelinkIntegrationMapping_ArchivesExistingMappings (seeds a mapping, delinks, asserts archived count + that it's no longer active) and TestDelinkIntegrationMapping_NotFound.

internal/service/entityintegrationmapping.go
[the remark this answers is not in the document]
Gursimar10: Wrapped the deletion loop in `s.DB.WithTx(ctx, ...)` and pass `txCtx` to each Delete, so all deletions commit or roll back atomically.

internal/api/dto/integration.go
Gursimar10: Kept the current fields as-is for now per your note; we can extend to accept `provider_entity_id` when needed.

internal/api/router.go
Gursimar10: Changed the route to `DELETE /integrations/link` (and updated the handler's Swagger `@Router` annotation to match).

PR #2024, opened 2026-06-18
feat(checkout): entity creation

Gursimar10:
## Summary by CodeRabbit

* **New Features**
  * Added hosted checkout support with authenticated endpoints to create and fetch checkouts (`POST /v1/checkout/session`, `GET /v1/checkout/session/:id`).
  * Checkout creation supports both `payment` and `setup` objectives with validation for success/cancel URLs and subscription parameters.
  * Webhook-driven completion now records and exposes checkout status and details, including hosted session outcomes.

* **Bug Fixes**
  * Webhook completion now correctly routes based on checkout-session metadata (payment vs setup), skips irrelevant events, and remains idempotent.
  * After successful invoice payments, eligible pending payment checkouts are completed automatically.


ent/schema/checkout.go
```diff
@@ -0,0 +1,87 @@
...
+			NotEmpty(),
+		field.String("entity_type").
+			SchemaType(map[string]string{"postgres": "varchar(50)"}).
+			GoType(types.CheckoutEntityType("")),
+		field.String("entity_id").
+			SchemaType(map[string]string{"postgres": "varchar(50)"}),
+		field.String("source_subscription_id").
+			SchemaType(map[string]string{"postgres": "varchar(50)"}).
+			Optional().Nillable(),
+		field.String("checkout_type").
```
omkar273: we will be removing this

ent/schema/checkout.go
```diff
@@ -0,0 +1,87 @@
...
+			SchemaType(map[string]string{"postgres": "varchar(50)"}).
+			GoType(types.CheckoutType("")),
+		field.String("objective").
```
omkar273: lets rename this to `checkout_action` instead of `checkout_type`

ent/schema/checkout.go
```diff
@@ -0,0 +1,87 @@
...
+			SchemaType(map[string]string{"postgres": "varchar(50)"}).
+			GoType(types.CheckoutObjective("")),
```
omkar273: rename this field to `mode`

ent/schema/checkout.go
```diff
@@ -0,0 +1,87 @@
...
+		field.Text("checkout_url").Optional().Nillable(),
+		field.Text("success_url").Optional().Nillable(),
+		field.Text("cancel_url").Optional().Nillable(),
+		field.JSON("configuration", map[string]interface{}{}).
+			Optional().
+			Comment("Reserved; deferred-operation payload (JSONB). Nil in v1."),
+		field.Time("expires_at"),
+		field.Time("completed_at").Optional().Nillable(),
+		field.Time("cancelled_at").Optional().Nillable(),
+		field.Text("error_message").Optional().Nillable(),
```
omkar273: lets call this `failure_message`

ent/schema/checkout.go
```diff
@@ -0,0 +1,87 @@
...
+		field.String("objective").
+			SchemaType(map[string]string{"postgres": "varchar(50)"}).
+			GoType(types.CheckoutObjective("")),
+		field.String("checkout_status").
+			SchemaType(map[string]string{"postgres": "varchar(50)"}).
+			GoType(types.CheckoutStatus("")).
+			Default(string(types.CheckoutStatusPending)),
+		field.Other("amount", decimal.Zero).
+			SchemaType(map[string]string{"postgres": "numeric(20,8)"}).
+			Optional(),
```
omkar273: we should have a default value 0

ent/schema/checkout.go
```diff
@@ -0,0 +1,87 @@
...
+		field.String("currency").Optional(),
+		field.String("provider").
+			SchemaType(map[string]string{"postgres": "varchar(50)"}),
```
omkar273: lets create a enum for this i.e. `flexprice`,`stripe` with default value `flexprice`

ent/schema/checkout.go
```diff
@@ -0,0 +1,87 @@
...
+			Annotations(entsql.IndexAnnotation{Where: "checkout_status = 'pending'"}),
+		// one in-flight upgrade per source subscription (NULL source rows unconstrained)
+		index.Fields("source_subscription_id").
+			Unique().
+			Annotations(entsql.IndexAnnotation{Where: "checkout_status = 'pending'"}),
+		// expiry sweep
+		index.Fields("expires_at").
+			Annotations(entsql.IndexAnnotation{Where: "checkout_status = 'pending'"}),
+	}
+}
```
omkar273, line 83:
```
// Indexes of the Subscription.
func (Subscription) Indexes() []ent.Index {
	return []ent.Index{
		// Common query patterns from repository layer
		index.Fields("tenant_id", "environment_id", "customer_id", "status").
			Annotations(entsql.IndexWhere("status = 'published'")),
		index.Fields("tenant_id", "environment_id", "plan_id", "status"),
		index.Fields("tenant_id", "environment_id", "subscription_status", "status"),
		// For billing period updates
		index.Fields("tenant_id", "environment_id", "current_period_end", "subscription_status", "status"),
		// Drives the plan-price sync's "which subs are behind?" lookup.
		index.Fields("tenant_id", "environment_id", "plan_id", "synced_price_sequence").
			Annotations(entsql.IndexWhere(
				"status = 'published' AND subscription_type IN ('standalone','delegated_invoicing','parent','grouped_invoicing')")),
	}
}
```

refactor indexes to use tenant id and environment to use partial indexes

internal/api/dto/checkout.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+	}
+	return nil
+}
+
+type CheckoutResponse struct {
+	ID          string `json:"id"`
+	Status      string `json:"status"`
+	CheckoutURL string `json:"checkout_url"`
+}
```
omkar273, line 65: lets return the enitre checkout entity instead

internal/api/v1/checkout.go
```diff
@@ -0,0 +1,85 @@
...
+// @Accept json
+// @Produce json
+// @Security ApiKeyAuth
+// @x-scope write
+// @Param request body dto.CreateCheckoutRequest true "Create checkout request"
+// @Success 201 {object} dto.CheckoutResponse
+// @Failure 400 {object} ierr.ErrorResponse "Invalid request"
+// @Failure 500 {object} ierr.ErrorResponse "Server error"
+// @Router /checkouts [post]
+func (h *CheckoutHandler) CreateCheckout(c *gin.Context) {
```
omkar273, line 37:
```suggestion
 /checkout/session [post]
```

internal/api/v1/checkout.go
```diff
@@ -0,0 +1,85 @@
...
+// @Accept json
+// @Produce json
+// @Security ApiKeyAuth
+// @x-scope read
+// @Param id path string true "Checkout ID"
+// @Success 200 {object} dto.CheckoutResponse
+// @Failure 400 {object} ierr.ErrorResponse "Invalid request"
+// @Failure 404 {object} ierr.ErrorResponse "Resource not found"
+// @Failure 500 {object} ierr.ErrorResponse "Server error"
+// @Router /checkouts/{id} [get]
```
omkar273:
```suggestion
 /checkout/session/{id} [post]
```

internal/api/router.go
```diff
@@ -345,6 +346,12 @@ func NewRouter(
 			}
 		}
 
+		checkout := v1Private.Group("/checkouts")
+		{
+			checkout.POST("", write("checkout", types.ActionWrite), handlers.Checkout.CreateCheckout)
+			checkout.GET("/:id", handlers.Checkout.GetCheckout)
+		}
```
omkar273:
```suggestion
 /checkout/session
```

internal/domain/checkout/model.go
```diff
@@ -0,0 +1,150 @@
...
+	EntityType types.CheckoutEntityType
+	EntityID   string
+
+	SourceSubscriptionID *string // upgrades only: old sub to cancel on completion
+
+	CheckoutType types.CheckoutType
+	Objective    types.CheckoutObjective
+	Status       types.CheckoutStatus
+
+	Amount   decimal.Decimal
```
omkar273: lets make amount nilable since in many flows it could ne nil i.e. save card

internal/domain/checkout/model.go
```diff
@@ -0,0 +1,150 @@
...
+	Currency string
+
+	Provider          string
```
omkar273: this should be enum

internal/domain/checkout/model.go
```diff
@@ -0,0 +1,150 @@
...
+	EnvironmentID string
+	types.BaseModel
+}
+
+// CheckoutConfiguration is reserved for a future "create nothing until paid" mode.
+// Kept generic to avoid a domain -> api/dto dependency; the typed deferred request
+// is marshaled into Payload by the service layer if/when that mode is built.
+type CheckoutConfiguration struct {
+	SaveCard bool            `json:"save_card,omitempty"`
+	Payload  json.RawMessage `json:"payload,omitempty"`
```
omkar273: lets make it typed struct with different keys for different actions 
```
subscription_create_params: CreateSubscriptionRequest

```

internal/types/checkout_test.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := tt.fn()
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("got err=%v, wantErr=%v", err, tt.wantErr)
+			}
+		})
+	}
+}
```
omkar273: lets remove this

internal/testutil/inmemory_checkout_store_test.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	got, err := store.ListPendingExpired(ctx, now)
+	if err != nil {
+		t.Fatalf("ListPendingExpired: %v", err)
+	}
+	if len(got) != 1 || got[0].ID != "chk_old" {
+		t.Fatalf("expected [chk_old], got %+v", got)
+	}
+}
```
omkar273, line 74: lets remove this

internal/repository/ent/checkout.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	if c.SuccessURL != nil {
+		builder.SetSuccessURL(*c.SuccessURL)
+	}
+	if c.CancelURL != nil {
+		builder.SetCancelURL(*c.CancelURL)
+	}
+	if configMap != nil {
+		builder.SetConfiguration(configMap)
+	}
```
omkar273, line 67: lets use nilable variants instead

```
		SetNillableSourceSubscriptionID(c.SourceSubscriptionID)

```

internal/repository/ent/checkout.go
```diff
@@ -0,0 +1,220 @@
...
+
+	span := StartRepositorySpan(ctx, "checkout", "update", map[string]interface{}{
+		"checkout_id": c.ID,
+		"status":      c.Status,
+	})
+	defer FinishSpan(span)
+
+	builder := client.Checkout.UpdateOneID(c.ID).
+		SetCheckoutStatus(c.Status).
+		SetUpdatedBy(c.UpdatedBy)
```
omkar273: lets also add tenant environment filter here

internal/repository/ent/checkout.go
```diff
@@ -0,0 +1,220 @@
...
+
+	span := StartRepositorySpan(ctx, "checkout", "list_pending_expired", nil)
+	defer FinishSpan(span)
+
+	entities, err := client.Checkout.Query().
+		Where(
+			entcheckout.CheckoutStatusEQ(types.CheckoutStatusPending),
+			entcheckout.ExpiresAtLT(cutoff),
+		).
+		All(ctx)
```
omkar273: lets add base filters like `pagination` , `status=published`

internal/service/payment_processor.go
```diff
@@ -1002,3 +1008,22 @@ func (p *paymentProcessor) handleIncompleteSubscriptionPayment(ctx context.Conte
...
+		return nil
+	}
+	chk, err := p.CheckoutRepo.GetPendingByEntity(ctx,
+		types.CheckoutEntityTypeSubscription, *inv.SubscriptionID, types.CheckoutObjectivePayment)
+	if err != nil {
+		return err
+	}
+	if chk == nil {
+		return nil
+	}
```
omkar273, line 1030: we should remove this ideally since value should be there when there is no error

internal/service/checkout.go
```diff
@@ -0,0 +1,346 @@
...
+	// driven by the existing payment-completion hook, not here.
+	Complete(ctx context.Context, checkoutID string) error
+	// Get returns a checkout by ID.
+	Get(ctx context.Context, id string) (*dto.CheckoutResponse, error)
+}
+
+type checkoutService struct {
+	ServiceParams
+	// providerFn resolves the checkout provider; overridable in tests.
+	providerFn func(ctx context.Context, provider string) (checkout.CheckoutProvider, error)
```
omkar273: we should create a enum for provider type instead of string

internal/domain/checkout/repository.go
```diff
@@ -0,0 +1,23 @@
...
+// Repository persists Checkout aggregates.
+type Repository interface {
+	Create(ctx context.Context, c *Checkout) error
+	Get(ctx context.Context, id string) (*Checkout, error)
+	Update(ctx context.Context, c *Checkout) error
+
+	// GetPendingByEntity returns the single pending checkout for a subject and
+	// objective (used by payment completion + dedupe). Returns nil, nil if none.
+	GetPendingByEntity(ctx context.Context, entityType types.CheckoutEntityType,
+		entityID string, objective types.CheckoutObjective) (*Checkout, error)
```
omkar273: we should create a struct for this instead it will be painful to extend it later

internal/domain/checkout/repository.go
```diff
@@ -0,0 +1,23 @@
...
+
+	// ListPendingExpired returns pending checkouts whose expires_at < cutoff.
+	ListPendingExpired(ctx context.Context, cutoff time.Time) ([]*Checkout, error)
```
omkar273: we should also add pagination so this

omkar273:
Payment-objective subscription never activates (see inline comment on `handler.go`)

Abandonment/expiry cleanup cron referenced in comments but not built; `checkout.session.expired` Stripe webhook not handled

`CheckoutResponse` is thin (missing `entity_id`, `amount`, `expires_at`); no list endpoint; `source_subscription_id` is an orphaned schema field since `subscription_change` was removed

internal/integration/stripe/webhook/handler.go
```diff
@@ -801,10 +801,30 @@ func (h *Handler) handleCheckoutSessionCompleted(ctx context.Context, event *str
...
 	flexpricePaymentID := checkoutSession.Metadata["flexprice_payment_id"]
-	if flexpricePaymentID == "" {
-		h.logger.Info(ctx, "no flexprice_payment_id found in checkout session metadata", "event_id", event.ID)
+	flexpriceCheckoutID := checkoutSession.Metadata["flexprice_checkout_id"]
+	if flexpricePaymentID == "" && flexpriceCheckoutID == "" {
+		h.logger.Info(ctx, "no flexprice metadata found in checkout session, skipping event",
+			"event_id", event.ID)
+		return nil
+	}
+	if flexpriceCheckoutID != "" {
```
omkar273: **🔴 Bug — payment-objective subscriptions never activate after this block**

`createPaymentSession` injects **both** `flexprice_checkout_id` and `flexprice_payment_id` into the Stripe session metadata. This block therefore always fires for payment sessions, calls `Complete()` (which only sets `checkout_status = completed` for the payment objective), and returns — `HandleFlexPriceCheckoutPayment` below is unreachable. The subscription stays `incomplete` indefinitely.

`payment_intent.succeeded` is also skipped for `PaymentLink` type, so there is no fallback.

**Suggested fix:** do not inject `flexprice_checkout_id` into payment-mode session metadata — keep the two paths cleanly separated by which key is present. Setup sessions carry `flexprice_checkout_id`; payment sessions carry only `flexprice_payment_id`.

internal/service/checkout.go
```diff
@@ -0,0 +1,346 @@
...
+		}
+		return s.CheckoutRepo.Create(txCtx, chk)
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// 5. Post-commit: open the hosted provider session (external call, outside the
+	// transaction). On failure the checkout is marked failed; the parked
+	// subscription/invoice are reaped by the abandonment cron.
```
omkar273: `ListPendingExpired` is implemented on the repo but has no caller — no cron handler, no Temporal workflow. Orphaned incomplete subscriptions and unpaid invoices will accumulate on expired/failed checkouts. Suggest tracking this as a follow-up issue before GA.

internal/api/dto/checkout.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		if r.Objective != types.CheckoutObjectivePayment && r.Objective != types.CheckoutObjectiveSetup {
+			return ierr.NewError("objective must be 'payment' or 'setup' for subscription_creation").
+				WithHint("Set objective to payment or setup").Mark(ierr.ErrValidation)
+		}
+	}
+	return nil
+}
+
+type CheckoutResponse struct {
```
omkar273, line 63: `CheckoutResponse` is missing a few fields callers will need: `entity_id` (so the frontend can poll the subscription), `amount` + `currency`, `expires_at` (for a countdown / retry prompt), and `objective`. Even if not all are needed in v1, `entity_id` and `expires_at` seem essential for a usable integration.

internal/api/dto/checkout.go
Gursimar10: Already done — `CheckoutResponseFromDomain` maps every field of the domain model (EntityID, Amount, Currency, Mode/objective, ExpiresAt, etc.) into `CheckoutResponse`.

internal/api/v1/checkout.go
Gursimar10: Already applied — annotation and router both use `/checkout/session`.

internal/repository/ent/checkout.go
Gursimar10: `SourceSubscriptionID` was removed in a prior refactor — the field and its setter no longer exist in the schema or domain model.

internal/api/dto/checkout.go
Gursimar10: All listed fields are already in `CheckoutResponse`: `entity_id`, `amount`, `currency`, `expires_at`, and `mode` (the renamed `objective`). The full mapping is in `CheckoutResponseFromDomain`.

internal/service/payment_processor.go
Gursimar10: `GetPendingByEntity` returns `(nil, nil)` when no matching checkout exists (not every subscription has a checkout — users can create subs directly). Removing the nil guard would panic on `chk.ID`. If the intent is that this path should always have a checkout, we'd need to change the repo to return `ErrNotFound` instead of `(nil, nil)`. Happy to do that if you want to tighten the contract — just want to confirm before touching the repo interface.

PR #2028, opened 2026-06-18
feat(kafka): optional second-cluster dual-write for AWS→GCP migration

ojasagarwal-flex:
## What

Adds **optional** dual-write to a second Kafka cluster, gated by config, to support the AWS→GCP migration. When the second cluster is configured, events are published to both clusters; when it isn't, behavior is unchanged (single-cluster, current production path).

## Why

During the AWS→GCP migration we need to mirror the event stream to the GCP-side Kafka without disrupting the live AWS pipeline. A config-gated dual-write lets us cut over incrementally and roll back instantly by flipping config.

## Changes

- `internal/config` — new dual-write config (second cluster connection), env-driven; `dualwrite_env_test.go` covers env wiring.
- `internal/kafka` — `event_publisher.go` / `producer.go` / `base.go` extended for an optional second publisher; `event_publisher_test.go` added.
- `internal/pubsub`, `internal/publisher`, `internal/pubsub/router` — thread the optional second-cluster path through.
- `cmd/server/main.go` — wire the optional second producer.

## Safety / rollback

- **Off by default** — no second cluster configured ⇒ identical to current behavior.
- No schema or consumer changes; purely an additive publish path.

## Testing

- `go build ./cmd/server/ ./internal/kafka/... ./internal/pubsub/... ./internal/publisher/... ./internal/config/...` ✅
- `go test ./internal/config/ ./internal/kafka/` ✅

## Note on base branch

Cut from `main` (not `develop`) intentionally — `develop` is mid-churn from the enterprise-package consolidation. Retarget to `develop` if the team prefers integrating there.

## Summary by CodeRabbit

* **New Features**
  * Added optional secondary cluster configuration, enabling simultaneous publishing to multiple Kafka clusters when enabled.

* **Tests**
  * Added test coverage for secondary cluster binding and multi-cluster publishing scenarios.

### DIFF
--- a/internal/kafka/event_publisher.go
+++ b/internal/kafka/event_publisher.go
@@ -14,18 +14,44 @@ import (
 	"github.com/flexprice/flexprice/internal/logger"
 )
 
-type EventPublisher struct {
-	producer *Producer
-	logger   *logger.Logger
-	config   *config.KafkaConfig
+// messagePublisher is the subset of the watermill publisher that EventPublisher depends
+// on. *Producer satisfies it; tests substitute a fake to exercise the fan-out without a broker.
+type messagePublisher interface {
+	Publish(topic string, messages ...*message.Message) error
 }
 
-func NewEventPublisher(producer *Producer, cfg *config.Configuration, logger *logger.Logger) *EventPublisher {
-	return &EventPublisher{
-		producer: producer,
-		logger:   logger,
-		config:   &cfg.Kafka,
+// EventPublisher publishes source events to Kafka. It always writes to the deployment's
+// local cluster (cfg.Kafka); during the AWS→GCP migration, if a second cluster is
+// configured (cfg.KafkaSecondary), it also writes there (presence-based dual-write). Writes
+// are independent: one cluster failing does not block the other. Ingest is fire-and-forget
+// at the service layer (CreateEvent logs and returns 202), so a failure here is logged
+// per-cluster and returned, not customer-facing. See infrastructure/docs/GCP-CUTOVER-STEPWISE.md.
+type EventPublisher struct {
+	primary      messagePublisher
+	primaryCfg   *config.KafkaConfig
+	secondary    messagePublisher
+	secondaryCfg *config.KafkaConfig
+	logger       *logger.Logger
+}
+
+func NewEventPublisher(primaryProducer *Producer, secondaryProducer *SecondaryProducer, cfg *config.Configuration, logger *logger.Logger) (*EventPublisher, error) {
+	if primaryProducer == nil {
+		return nil, fmt.Errorf("kafka producer is not initialized")
 	}
+	ep := &EventPublisher{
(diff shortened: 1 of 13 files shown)


agrim123 (commented):
How about if we just use []KafkaConsumer in event publisher in ordered way. That way the first will always be primary and rest can fired off as goroutines

ojasagarwal-flex:
> How about if we just use []KafkaConsumer in event publisher in ordered way. That way the first will always be primary and rest can fired off as goroutines

slice of consumers is i think not needed, we are gonna have two kafka brokers only, creating slice would be an overhead. Also we can make the secondary write async using library config. goroutine can remain unbounded

internal/kafka/event_publisher.go
```diff
@@ -36,44 +62,62 @@ func (p *EventPublisher) Publish(ctx context.Context, event *events.Event) error
...
 	partitionKey := event.TenantID
 	if event.ExternalCustomerID != "" {
 		partitionKey = fmt.Sprintf("%s:%s", event.TenantID, event.ExternalCustomerID)
 	}
+
+	// Write to each enabled cluster independently — one failing must not block the other.
+	// A fresh message per cluster (the watermill publisher consumes the message it is handed);
+	// same event.ID + payload everywhere keeps the streams dedup-identical.
+	var firstErr error
+	if p.primary != nil {
```
agrim123: Primary can never be null. And it is fails, the request has to abort immediately.
If we are not respeecting it then there is no point in calling it primary, array is better in that case.

internal/pubsub/kafka/base.go
```diff
@@ -17,11 +17,15 @@ import (
 )
 
 func GetSaramaConfig(cfg *config.Configuration) *sarama.Config {
+	// This stack (webhook pub/sub, DLQ, consumers) operates on the deployment's local
+	// cluster — the one it reads from — not necessarily the AWS write target.
+	kc := &cfg.Kafka
```
agrim123: why changing this? This is leading to function signature modification across the code

agrim123: Maybe I am understanding this wrong, please test it when both producers have different auth mechanisms.

agrim123 (changes requested):
One more thing apart from inline comments: have we verified that consumer config is completely independent of producer configs?
Example: GCP primary producer doesn't conincide with AWS or vice versa in auth

ojasagarwal-flex:
> One more thing apart from inline comments: have we verified that consumer config is completely independent of producer configs? Example: GCP primary producer doesn't conincide with AWS or vice versa in auth

yes i have verified this. Producer configs are different. Consumer is for its respective deplopyments

agrim123 (approved):
LGTM, test cases to cover
1. single producer
2. two producers same auth
3. two producer different auth

agrim123 (approved):

PR #2029, opened 2026-06-18
feat(secret): add support for service account name in list api keys response

subratsahilgupta:
## Summary by CodeRabbit

* **New Features**
  * Secret-related API responses now include user ownership metadata, including `user_id` and `service_account_name`.
  * When listing API keys, the associated service account name is now shown.

* **Bug Fixes**
  * Added validation to prevent creating service accounts with `super_admin` included alongside other roles.

### DIFF
--- a/internal/ee/service/secret.go
+++ b/internal/ee/service/secret.go
@@ -190,11 +190,28 @@ func (s *secretService) ListAPIKeys(ctx context.Context, filter *types.SecretFil
 	}
 
 	items := dto.ToSecretResponseList(secrets)
+
+	serviceAccountNameByID := make(map[string]string)
 	for _, item := range items {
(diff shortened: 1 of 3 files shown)


internal/ee/service/secret.go
```diff
@@ -189,8 +189,18 @@ func (s *secretService) ListAPIKeys(ctx context.Context, filter *types.SecretFil
...
 
+	items := dto.ToSecretResponseList(secrets)
+	for _, item := range items {
+		if item.UserType == types.UserTypeServiceAccount && item.UserID != "" {
+			u, err := s.userRepo.GetByID(ctx, item.UserID)
+			if err == nil {
+				item.ServiceAccountName = u.Name
+			}
+		}
+	}
```
omkar273, line 217: instead of this make a list of all user ids and use list method insetad of api call for each user

PR #2034, opened 2026-06-19
feat(logging): add WebhookLoggingConfig for webhook request persistence 

Gursimar10:
## Release Notes

- **New Features**
  - Added configurable webhook request logging on the public `/v1/webhooks` endpoints to capture and persist provider/method/path, request ID, headers, and body for selected tenants and environments.
  - Oversized payloads (over 1 MiB) are logged but not persisted.

- **Updates**
  - Extended Zoho invoice synchronization to include `invoice_number` and store it in integration mapping metadata.

- **Tests**
  - Added unit tests for provider extraction, allowlist matching, and middleware request-body buffering.


ent/schema/incoming_webhook_event.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"entgo.io/ent"
+	"entgo.io/ent/schema/field"
+	"entgo.io/ent/schema/index"
+)
+
+// WebhookRequest holds the schema for inbound webhook request audit logs.
+type WebhookRequest struct {
+	ent.Schema
+}
```
omkar273, line 13: Let's give a better name for this table, `incoming_webhook_events` or can think what can be a better alternative

ent/schema/webhook_request.go
```diff
@@ -0,0 +1,72 @@
...
+	ent.Schema
+}
+
+func (WebhookRequest) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("id").
+			SchemaType(map[string]string{"postgres": "varchar(50)"}).
+			Unique().
+			Immutable(),
+		field.String("tenant_id").
```
omkar273: lets use baseMixin here instead

PR #2088, opened 2026-06-25
fix(config): bind FLEXPRICE_CHECKOUT_BASE_URL env var

ojasagarwal-flex:
## What

Adds an explicit `v.BindEnv("checkout.base_url", "FLEXPRICE_CHECKOUT_BASE_URL")`.

## Why

`CheckoutConfig.BaseURL` is delivered to GKE deployments only as an env var (Helm `extraEnv`) and is **not** rendered into the mounted ConfigMap. viper's `Unmarshal` does not consult `AutomaticEnv` for keys absent from the config file, so without this bind `checkout.base_url` resolves to empty and checkout links break. (No boot failure — the parent field is `validate:"omitempty"` — but the feature silently doesn't get its base URL.)

Same class/pattern as the existing `auth.secret`, `redis.*`, `svix`, and `clickhouse.*` binds.

## Companion

Staging value delivered via infra `extraEnv` (`FLEXPRICE_CHECKOUT_BASE_URL=https://admin-dev.flexprice.io`). Takes effect once a develop image with this bind is deployed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration loading so the checkout base URL can now be set from an environment variable even when it is missing from the main configuration file.
  * This makes checkout setup more reliable across environments where configuration values may be injected externally.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -723,6 +723,11 @@ func NewConfig() (*Configuration, error) {
 	// NOTE: auth.api_key.keys is intentionally NOT bound here because the env var is a
 	// JSON string but Viper/mapstructure expects a map. It is handled manually in Step 6.
 
+	// checkout.base_url is delivered only as an env var (extraEnv) and is absent from the
+	// Helm-rendered ConfigMap, so Unmarshal would otherwise leave it empty. Bind it so the
+	// FLEXPRICE_CHECKOUT_BASE_URL value is actually read.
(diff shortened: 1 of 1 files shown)


ojasagarwal-flex:
Superseded: switching to the standard pattern — render checkout.base_url in configmap.yaml from a chart value (like webhook/email/billing), which the app reads from the config file. The BindEnv approach is unnecessary for a configmap-rendered key.

PR #2132, opened 2026-06-30
feat(kafka): add dlq on consumer group which ingest from events

Gursimar10:
## Summary by CodeRabbit

* **New Features**
  * Added per-consumer-group Kafka DLQ topic overrides (`topic_dlq`) across multiple consumers, including lazy variants, so each consumer can route failures to its own dead-letter topic.
  * If `topic_dlq` is unset, consumers fall back to the shared Kafka DLQ topic.

* **Bug Fixes**
  * Improved post-retry “poison” handling by routing failed messages to the explicitly configured DLQ topic when available.
  * Updated webhook and integration event consumers to participate in the same explicit DLQ routing behavior.


internal/pubsub/router/router.go
```diff
@@ -115,12 +79,15 @@ func createDLQPublisher(cfg *config.Configuration, logger *logger.Logger) (messa
...
-// AddNoPublishHandler adds a handler that doesn't publish messages
+// AddNoPublishHandler adds a handler that doesn't publish messages.
+// topicDLQ overrides the global DLQ topic for this handler; pass "" to use
+// the global kafka.topic_dlq fallback.
 func (r *Router) AddNoPublishHandler(
 	handlerName string,
 	topicName string,
 	subscriber message.Subscriber,
 	handlerFunc func(msg *message.Message) error,
+	topicDLQ string,
```
DiabolusGX: `topic` and `topicDLQ` next to each other for better readability
```suggestion
    handlerName string,
    topicName string,
    topicDLQ string,
    subscriber message.Subscriber,
    handlerFunc func(msg *message.Message) error,
```

PR #2146, opened 2026-07-01
feat(cache): repo caching

Gursimar10:
## Summary by CodeRabbit

* **New Features**
  * Added caching for several frequently used records, improving read performance for environments, users, groups, and other core data.

* **Bug Fixes**
  * Improved cache reliability by using clearer cache keys, safer cache decoding, and better handling of cache misses and backend errors.
  * Reduced stale data by invalidating cached records after create, update, and delete actions.
  * Adjusted cache lifetimes to better balance freshness and performance.


internal/cache/inmemory.go
```diff
@@ -62,55 +62,84 @@ func GetInMemoryCache() InMemoryCache {
 }
 
 // Get retrieves a value from the cache
-func (c *inMemoryCache) Get(_ context.Context, key string) (interface{}, bool) {
+func (c *inMemoryCache) Get(ctx context.Context, key string) (interface{}, bool) {
 	if c == nil || !c.IsEnabled() {
 		return nil, false
 	}
-	return c.cache.Get(key)
+	if ctx.Err() != nil {
```
DiabolusGX: I don't think we need this check, any all methods wherever added in this PR

internal/cache/inmemory.go
```diff
@@ -62,55 +62,84 @@ func GetInMemoryCache() InMemoryCache {
...
+		return nil, false
+	}
+	value, found := c.cache.Get(key)
+	if found {
+		RecordHit(ctx, entityFromKey(key), SourceInMemory)
+	} else {
+		RecordMiss(ctx, entityFromKey(key), SourceInMemory)
```
DiabolusGX: either make a common function, `recordCacheHit` which will set `hit = true/false` or just remove both, we can rely on redis cache hit rate metric

internal/cache/inmemory.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // DefaultExpiration is the default expiration time for cache entries
-const DefaultExpiration = 30 * time.Minute
+const DefaultExpiration = 2 * time.Minute
```
DiabolusGX, line 14: this is not being used anywhere right? Should we remove it?

internal/cache/observe.go
```diff
@@ -0,0 +1,66 @@
...
+	sp.AddEvent(name, trace.WithAttributes(
+		attribute.String("cache.entity", entity),
+		attribute.String("cache.source", source),
+	))
+}
+
+// RecordLookup records a cache lookup outcome as an OTel span event, emitting a
+// hit event when hit is true and a miss event otherwise. entity is the domain
+// entity name (e.g. "customer"), source is "inmemory" or "redis".
+func RecordLookup(ctx context.Context, entity, source string, hit bool) {
```
DiabolusGX:
```suggestion
func recordLookup(ctx context.Context, entity, source string, hit bool) {
```

internal/cache/inmemory.go
```diff
@@ -83,31 +85,38 @@ func (c *inMemoryCache) ForceCacheSet(ctx context.Context, key string, value int
 	c.cache.Set(key, value, expiration)
 }
 
-func (c *inMemoryCache) ForceCacheDelete(_ context.Context, key string) {
+func (c *inMemoryCache) ForceCacheDelete(ctx context.Context, key string) {
```
DiabolusGX: there's no `recordDelete` in this?

internal/cache/observe.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// Cache source labels used in observability events.
+const (
+	SourceInMemory = "inmemory"
+	SourceRedis    = "redis"
+)
+
+// recordEvent adds a cache observability event to the active span, if any.
+func recordEvent(ctx context.Context, name, entity, source string) {
```
DiabolusGX: I'd say we can directly use this method and don't need wrappers like `recordSet` or `recordDelete`.
See how we're currently using `StartRepositorySpan` in ent repos.

DiabolusGX: We can return `trace` take the same trace in input of helpers (like `RecordLookup`) to mark it as hit/miss.
We're currently starting new trace just to mark another trace hit/miss. It will cost a lot more plus the single trace will not have all of its own details

internal/repository/factory.go
```diff
@@ -152,39 +152,39 @@ func NewEntitlementRepository(p RepositoryParams) entitlement.Repository {
...
+	return entRepo.NewSecretRepository(p.EntClient, p.Logger, p.RedisCache)
 }
 
 func NewCreditGrantRepository(p RepositoryParams) creditgrant.Repository {
-	return entRepo.NewCreditGrantRepository(p.EntClient, p.Logger, p.InMemoryCache)
+	return entRepo.NewCreditGrantRepository(p.EntClient, p.Logger, p.RedisCache)
 }
 
 func NewCostsheetRepository(p RepositoryParams) costsheet.Repository {
 	return entRepo.NewCostsheetRepository(p.EntClient, p.Logger, p.InMemoryCache)
```
DiabolusGX: remove `p.InMemoryCache` if not being used from all entities

internal/cache/inmemory.go
Gursimar10: it is being used during InitializeInMemoryCache

DiabolusGX (approved):

PR #2157, opened 2026-07-03
feat(meter-usage): Use redis lock to prevent duplicates

DiabolusGX:
## Summary by CodeRabbit

* **New Features**
  * Added `redis_deduplication_enabled` (default: enabled) to prevent duplicate meter-usage event processing via distributed locking.
  * Plan price sync now uses the same distributed lock mechanism to avoid concurrent sync runs for the same plan.
* **Improvements**
  * Cache keys are now context-aware, and more cache reads/writes are backed by Redis for more consistent behavior across the service.
* **Documentation**
  * Updated release messaging to better reflect lock acquisition outcomes versus “already in progress” scenarios.


internal/ee/service/meter_usage_tracking.go
```diff
@@ -286,6 +288,21 @@ func (s *meterUsageTrackingService) getMetersForEvent(ctx context.Context, event
...
 func (s *meterUsageTrackingService) processEvent(ctx context.Context, event *events.Event) error {
+	// Step 0: Check if the event is already processed
+	if s.Config.MeterUsageTracking.RedisDeduplicationEnabled &&
+		s.ServiceParams.RedisCache != nil {
+		eventId := event.ID
+		cacheKey := cache.GenerateKey(cache.PrefixEvent, eventId)
+		locked, err := s.ServiceParams.RedisCache.AcquireLock(ctx, cacheKey, eventDeduplicationLockTTL)
+		if err != nil {
+			return fmt.Errorf("failed to check if event is already processed: %w", err)
+		}
```
nkmishra1997, line 299: should clear lock on error so that the retry event does not skip it

PR #2161, opened 2026-07-03
fix(tracing): make DB spans visible as Database Calls in OTEL + per-statement Postgres/ClickHouse spans

nkmishra1997:
## Problem

`FLEXPRICE_OTEL_TRACES_STORAGE_SPANS_ENABLED=true` was enabled on production (`flexprice-api-v4`, ap-south-1) but no DB spans appeared in SigNoz's trace waterfall highlighting or the **Database Calls** tab.

Investigation showed the spans (`postgres.transaction`, `clickhouse.*`) **were** being exported, but as `SpanKindInternal` with no `db.system` attribute. SigNoz's Database Calls tab only counts spans with `spanKind = Client` **and** a non-empty `db.system`, so they were invisible where operators look for DB activity.

Additionally, Postgres coverage was transaction-only, several ClickHouse operations (`query`, `exec`, `prepare_batch`, `async_insert`, `batch_send`) had their instrumentation commented out, and the repository layer (~49 files across `internal/repository/ent` and `internal/repository/clickhouse`) already calls `StartRepositorySpan` around essentially every read/write — but both implementations were no-op stubs, so none of those existing call sites ever emitted anything.

## Changes

**`internal/tracing/tracing.go`**
- `StartDBSpan` / `StartClickHouseSpan` / `StartRepositorySpan` now start spans with `SpanKindClient` and the OTel `db.system` semconv attribute (`postgresql` / `clickhouse`) via a shared `startStorageSpan` helper.
- `startStorageSpan` itself is gated on `otel.traces.storage_spans_enabled`, so every storage span obeys the one flag regardless of call path (per review).
- `StartRepositorySpan` takes an explicit `dbSystem` argument.

**`internal/repository/ent/helper.go`, `internal/repository/clickhouse/helper.go` — the existing repo-layer spans, now live**
- The previously no-op `StartRepositorySpan` free functions are wired to the real `tracing.Service` via a package-level `SetTracingService` setter, so no repository constructor changes. Every existing call site (~49 files) now emits a real `repository.<entity>.<operation>` span tagged `SpanKindClient` + `db.system` — this is what provides per-read/write Postgres and ClickHouse coverage.
- `internal/repository/factory.go` adds `InitTracing(tracingSvc)`; called via `fx.Invoke` in `cmd/server/main.go`.
- `internal/ee/service/event.go`'s two direct callers pass `db.system=clickhouse` explicitly.

**`internal/postgres/client.go`**
- The `TracingClient` `IClient` wrapper (`monitoring.go`) is deleted; the `postgres.transaction` span and reader/writer routing tags (`db.endpoint`, `db.resolved_target`) now live directly on `Client.WithTx`/`Reader`/`Writer` with nil-safe tracing checks. `NewEntClients` keeps its original `(config, logger)` signature — no script changes.
- An earlier revision added a custom traced Ent `dialect.Driver` for per-SQL-statement spans; it was dropped in review as redundant with the repository-layer spans (see discussion). If raw-SQL spans are ever needed, an off-the-shelf `database/sql` instrumentation lib is the better route.

**`internal/clickhouse/clickhouse.go`**
- Enabled the previously commented-out spans (`query`, `exec`, `prepare_batch`, `async_insert`, `batch_send`); standardized the query-text attribute to `db.statement`; `batch_send` now parents to the PrepareBatch context instead of being dropped.
- All traced methods record error status on failure so failed DB calls no longer render as successful spans (per review).

All spans remain gated on `FLEXPRICE_OTEL_TRACES_STORAGE_SPANS_ENABLED` (default false); at default everything above is a no-op.

## Known limitation

Repository-layer call sites discard the child context `StartRepositorySpan` returns, so repo spans render as siblings of the request span rather than nesting lower-level spans (e.g. ClickHouse query spans) under them. Fixing that means threading a reassigned context through ~150 call sites across 49 files — deferred. They still appear correctly in the Database Calls tab and search.

## Verification

- `go build ./cmd/... ./internal/... ./scripts/...`, `go vet`, and `gofmt` clean (excluding a pre-existing vet issue in the generated `api/custom/go` SDK).
- `go test`: 1848 tests pass across the 9 affected packages.
- Verified against production: flag confirmed set on task def `web_api:570`; SigNoz confirmed spans arriving with `kind=Internal`, `db.system=''` (the root cause), including the previously-no-op `repository.GetUsageByMeter.*` spans.

## Note for reviewers

Span volume increases from ~1 span per transaction to ~1 span per repository method call plus per-query ClickHouse spans. The flag can be flipped off instantly via env if ingest cost is a concern.

## Summary by CodeRabbit

* **New Features**
  * Repository, database, and ClickHouse operations now emit richer tracing data for better request visibility and diagnostics.
  * Transaction activity can now be traced more consistently when storage spans are enabled.

* **Bug Fixes**
  * Improved error reporting in traced database and ClickHouse operations so failed calls are marked correctly.
  * Connection, batch, and async insert flows now finish spans more reliably, reducing missing trace data.


internal/postgres/traced_driver.go
```diff
@@ -0,0 +1,129 @@
...
+// Span creation is gated per call on otel.traces.storage_spans_enabled
+// (env: FLEXPRICE_OTEL_TRACES_STORAGE_SPANS_ENABLED); at default (false) the
+// wrapper is a pure passthrough.
+type tracedDriver struct {
+	dialect.Driver
+	tracing  *tracing.Service
+	endpoint string // "writer" or "reader"
+}
+
+func newTracedDriver(drv dialect.Driver, tracingSvc *tracing.Service, endpoint string) dialect.Driver {
```
nkmishra1997: is this traced driver still required and are we still using it? if no remove if yes then can we try to do without it?

nkmishra1997: Removed in the latest commit. It was live (it was the source of the per-SQL-statement `postgres.select`/`postgres.insert` spans), but with the repository-layer spans now wired — every read/write already emits `repository.<entity>.<operation>` as a DB client span — the statement-level granularity wasn't worth a custom driver type plus the extra span volume. `traced_driver.go` is deleted and `NewEntClients` is back to its original `(config, logger)` signature, so all the script call-site changes are reverted too. If we ever want raw-SQL spans later, the cleaner route would be an off-the-shelf `database/sql` instrumentation lib (e.g. otelsql) rather than a hand-rolled Ent driver wrapper.

DiabolusGX (approved):

PR #2200, opened 2026-07-08
feat(tracing): emit db.resolved_target on Kafka consumer workloads

ojasagarwal-flex:
## Problem

The reader/writer Postgres router records the `db.resolved_target` span tag (`reader` / `writer` / `writer_pinned` / `writer_via_tx` / `writer_forced`) **only on the span already present in the context** (`GetSpanFromContext`) — it does not create one:

```go
// internal/postgres/client.go
if span := c.tracing.GetSpanFromContext(ctx); span != nil {
    span.SetTag("db.resolved_target", target)   // skipped entirely if span == nil
}
```

- **HTTP API:** otelgin creates a per-request server span → tag lands → visible in SigNoz.
- **Consumers:** each handler processes its message on a bare `context.Background()` with **no span**, so `GetSpanFromContext` returns nil and the tag is silently dropped. DB routing is invisible on the consumer workloads — you can't tell whether consumer reads hit the replica, the writer, or resolved to `writer_pinned` after a write.

## Fix

Start a per-message root span on each consumer handler, mirroring the package-global pattern the ent/clickhouse repo helpers already use (`StartRepositorySpan`), so no `*tracing.Service` has to be threaded through 9 constructors:

- **`internal/tracing/global.go`** — `SetDefaultService` + `StartConsumerSpan(ctx, name)`. Nil-safe: returns `(nil, ctx)` when tracing is disabled/unwired, and `*Span` methods are nil-safe.
- **`InitTracing`** — wires the default service next to the existing ent/clickhouse wiring (one call site).
- **9 `ee/service` consumer handlers** — start the span immediately after each builds its per-message context, so downstream DB spans resolve their target onto a live span. Brings consumers to parity with the API.

### Special case: `onboarding.go`
Its context is handed to a goroutine (`go s.generateEvents(bgCtx, …)`) that outlives the handler. A function-scope `defer span.Finish()` would end the span before the async DB writes run and drop the tag, so the span is finished **inside** the goroutine instead:
```go
go func() {
    defer span.Finish()
    s.generateEvents(bgCtx, &eventMsg)
}()
```

## Scope / risk

Observability-only. No behavioral change, and a complete no-op when `FLEXPRICE_OTEL_TRACES_*` is disabled. `go build ./internal/...` and `go vet ./internal/ee/service/` are green; gofmt-clean.

## Verify after deploy
In SigNoz, filter consumer traces (e.g. span name `kafka.consume.event_post_processing`) and confirm `db.resolved_target` is now populated — expect `writer_pinned` on reads that follow a write in the same message (read-your-writes via the writer pin), and `reader` on independent reads.

## Summary by CodeRabbit

* **Enhancements**
  * Improved message processing by preserving context across event, usage, webhook, and integration workflows.
  * Added consistent tracing for consumed messages, including success and error status reporting.
  * Improved context propagation for tenant and environment information during downstream processing.

* **Bug Fixes**
  * Improved reliability of context-dependent routing, processing, and monitoring across asynchronous message handling.


internal/ee/service/onboarding.go
```diff
@@ -267,8 +267,17 @@ func (s *onboardingService) processMessage(msg *message.Message) error {
...
+	// goroutine that outlives this handler, so the span is finished when that
+	// goroutine completes (not via defer here) or the tag would be dropped.
+	span, bgCtx := kafka.StartConsumerSpan(bgCtx, "onboarding")
+
 	// Start a goroutine to generate events at a rate of 1 per second
-	go s.generateEvents(bgCtx, &eventMsg)
+	go func() {
+		defer span.Finish()
+		s.generateEvents(bgCtx, &eventMsg)
+	}()
```
nkmishra1997: can make this sync now

nkmishra1997 (approved):

PR #2236, opened 2026-07-13
feat(webhooks): add event.rejected webhook for unmetered events (FLE-940)

ojasagarwal-flex:
## What & why
Sarvam (and any customer) has no visibility today when an ingested usage event matches **no meter** — it's silently dropped in the meter-usage consumer, so misconfigured/typo'd event names quietly lose usage. This adds an **`event.unmatched`** webhook for those events.

Fires when an event produces **no meter usage**, covering both drop points in `meter_usage_tracking.go`:
- no meter is registered for the event name, **or**
- meters exist for the name but filters/quantity excluded all of them (previously not even logged).

## Design
- **Opt-in**: `meter_usage_tracking.unmatched_event_webhook_enabled` (default `false`).
- **Throttled**: at most once per configurable window (`unmatched_event_webhook_window`, default `10m`) per `(tenant, env, event_name)`, via a Redis lock keyed by event name — distinct from the existing event-ID dedup lock. No Redis ⇒ fail-open (publishes every unmatched event rather than dropping the signal).
- **Best-effort**: publish failures are logged, never returned — can't block or fail meter processing.
- **Pass-through payload builder**: the event snapshot is in-hand at publish time, so no entity re-fetch / ClickHouse round-trip (unlike the feature/customer builders).
- **Audience filtering is downstream**: only tenants subscribed via Svix / native config receive it; the producer publishes unconditionally.

## Files
- `types/webhook.go`, `types/entity.go` — `event.unmatched` name, reason, `SystemEntityTypeEvent`
- `webhook/dto/event.go`, `webhook/payload/event.go`, `webhook/payload/factory.go` — DTO + builder + registration
- `config/config.go`, `config/config.yaml` — two config fields (off by default)
- `ee/service/meter_usage_tracking.go` — `publishUnmatchedEventWebhook` helper + single call site
- `api/v1/webhook_internal.go` — OpenAPI doc stub

## Testing
- Unit tests: builder round-trip + invalid input; helper disabled-noop, enabled-publishes, per-event-name throttle, no-locker fail-open.
- `go build ./internal/...`, `go vet`, `gofmt` clean.

## Follow-ups (not in this PR)
- Not yet driven end-to-end through Kafka→Svix (verified via unit tests + build).
- `make swagger` not run — doc stub is in place but the spec/SDKs aren't regenerated.

FLE-940

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in `event.rejected` webhook for ingested events that produce no meter usage.
  * Webhook payloads now include the event details plus a rejection reason (`no_meter_for_event_name` or `no_matching_meter`).
  * Added configurable throttling to limit notifications (default: once per event name every 10 minutes).
  * Added configuration to enable the webhook and set the throttling window.
* **Bug Fixes**
  * Rejected-event webhooks are now emitted when no matching meter-derived usage is found.
* **Tests**
  * Added coverage for disabled, successful, throttled, and fail-open webhook delivery scenarios.


ojasagarwal-flex:
### Update
- Webhook event renamed **`event.unmatched` → `event.rejected`** (the `reason` field carries the specifics: `no_meter_for_event_name` / `no_matching_meter`).
- Both reasons retained; single publish call site in `processEvent`.
- Tests now use `testutil.NewInMemoryRedisLocker` (SetNX+TTL) instead of an ad-hoc fake.

⚠️ **This branch also now contains a merge of `develop`** (182 files) that was in progress in the worktree — so this PR is no longer FLE-940-only. If you'd prefer a clean, review-friendly PR, I can rebase the FLE-940 commits onto latest `develop` and drop the merge noise. Also note: the merge's conflict resolution had dropped the FLE-940 edits in `meter_usage_tracking.go`; they've been re-applied and verified (build + tests green).

DiabolusGX (approved):

PR #2250, opened 2026-07-14
fix(rbac): enforce permission denial for service accounts

agrim123:
## Summary
- `RequirePermission` (`internal/rest/middleware/permission.go`) computed the correct RBAC deny decision for service-account API keys lacking a required role, but only logged it — no `AbortWithStatusJSON`/`return`, so the request always proceeded to `c.Next()` regardless of the check's result.
- Live-verified on staging: a key scoped to `roles: ["event_ingestor", "event_reader"]` could successfully `POST /v1/customers` (got `201`, customer actually created) and `GET /v1/plans` (got `200`) — both outside its declared role scope, both should have been `403`.
- Everything upstream of this bug is correct: route registration (`write(types.EntityCustomer, types.ActionWrite)`), API-key role loading into context (`auth.go`), and the RBAC role→permission mapping in `roles.json` all work as intended. The only defect is the missing abort in the deny branch.

## Fix
Added `c.AbortWithStatusJSON(http.StatusForbidden, ...)` + `return` in the deny branch, matching the pattern already used for the suspended-tenant check earlier in the same function.

Also corrected a pre-existing test in `tenant_access_test.go` that had asserted the buggy fall-through as expected behavior (`"service account without required role is logged but allowed"` expecting `200`) — updated to assert the correct `403`.

## Test plan
- [x] New test `TestRequirePermission_DeniesServiceAccountWithoutRole` reproduces the exact staging PoC (event_ingestor/event_reader vs customer:write) against the real `roles.json` config — fails pre-fix (200), passes post-fix (403). Verified via `git stash` on just the fix file.
- [x] New test `TestRequirePermission_AllowsServiceAccountWithRole` confirms the allow path still works (uses the real `super_admin` wildcard role).
- [x] `go test ./internal/rest/middleware/... ./internal/api/... ./internal/ee/service/... ./internal/types/...` — all green.
- [ ] Manual re-verification against staging recommended before/after merge (mint a scoped key, confirm `POST /v1/customers` now returns 403).

## Summary by CodeRabbit

- **Bug Fixes**
  - Service accounts lacking required RBAC permissions are now reliably denied with **HTTP 403** and a clear **“insufficient permissions”** JSON message.
  - Requests gated by RBAC now stop immediately on denial.
  - API key authentication now propagates the correct user identifier after a successful key match.

- **Tests**
  - Added and updated middleware tests to verify both allowed and forbidden service-account scenarios, including **403** response assertions.

### DIFF
--- a/internal/rest/middleware/auth.go
+++ b/internal/rest/middleware/auth.go
@@ -30,7 +30,7 @@ func validateAPIKey(ctx context.Context, cfg *config.Configuration, secretServic
 		secret, err := secretService.VerifyAPIKey(ctx, apiKey)
 		if err == nil && secret != nil {
 			// Return roles from the secret for RBAC permission checks
-			return secret.TenantID, secret.CreatedBy, secret.EnvironmentID, secret.UserType, secret.Roles, true
+			return secret.TenantID, secret.UserID, secret.EnvironmentID, secret.UserType, secret.Roles, true
 		}
 	}
 
--- a/internal/rest/middleware/permission.go
+++ b/internal/rest/middleware/permission.go
@@ -42,7 +42,7 @@ func (pm *PermissionMiddleware) RequirePermission(entity types.Entity, action ty
 		if types.IsServiceAccount(ctx) {
 			roles := types.GetRoles(ctx)
 			if !pm.rbacService.HasPermission(roles, string(entity), string(action)) {
-				pm.logger.Info(ctx, "service account access refrained due to insufficient RBAC roles",
+				pm.logger.Info(ctx, "service account access denied due to insufficient RBAC roles",
 					"user_id", types.GetUserID(ctx),
 					"tenant_id", types.GetTenantID(ctx),
 					"environment_id", types.GetEnvironmentID(ctx),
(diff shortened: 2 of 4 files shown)


agrim123:
## Blocked for merge — needs tenant API key audit/migration first

Team flagged a real rollout risk: some tenants may already be using service-account API keys whose declared \`roles\` don't cover everything they're actually calling (e.g. a key scoped \`event_ingestor\`/\`event_reader\` that's in practice been hitting non-event endpoints). Enforcing the RBAC check as written here would break those integrations immediately on merge, with no migration path.

Checked prod (read-only, \`prod_flexprice\` on \`flexprice-prod-pg-dest\`):

| user_type | roles | active keys |
|---|---|---|
| service_account | `["event_ingestor", "event_reader"]` | 15 |
| service_account | `["event_ingestor"]` | 11 |
| service_account | `["event_reader", "event_ingestor"]` | 6 |
| service_account | `["super_admin"]` | 6 (unaffected — wildcard) |

**32 active service-account keys are narrowly scoped** (`event_ingestor`/`event_reader`) and would be subject to this fix's enforcement.

Could not determine which of these are still live/in-use from the DB alone — `last_used_at` is unpopulated for all 186 private_key secrets in prod (dead column, not just stale). Confirming real usage needs a request-log cross-reference (Cloud Logging, by key ID, against non-event write endpoints) that hasn't been done yet.

**Action needed before this merges:** identify which of these 32 keys are actively calling out-of-scope endpoints, and migrate/re-issue those tenants' keys with correct roles first — otherwise this fix will break them the moment it lands.

nkmishra1997:
@agrim123 @subratsahilgupta lets validate usage in logs here and close this!

subratsahilgupta:
@DiabolusGX bhai merge please

DiabolusGX (approved):

PR #2251, opened 2026-07-14
fix(whop): verify webhook signature before processing events

agrim123:
## Summary
- `HandleWhopWebhook` (`internal/api/v1/webhook.go`) had no signature verification at all — unlike every other payment provider integration in this file (Stripe, HubSpot, Razorpay, Chargebee, QuickBooks, Nomod, Moyasar, Paddle, Zoho), which all verify a signature/secret before processing.
- Live-verified on staging: `POST /v1/webhooks/whop/{tenant}/{env}` with no auth header and no signature header, carrying a fully forged payment payload, returned `200` and was handed to `HandleWebhookEvent`. Control test: the identical no-signature attempt against `POST /v1/webhooks/stripe/...` correctly returned `400 {"error":"Missing Stripe-Signature header"}`.

## Fix
- Added `WebhookSecret` (optional, encrypted) to `WhopConnectionMetadata` — not required in `Validate()`, so existing Whop connections without one keep working for non-webhook flows (product sync, invoicing).
- Added `VerifyWebhookSignature(ctx, payload, signature) error` to `WhopClient`, HMAC-SHA256 over the raw body against the decrypted `webhook_secret`, `X-Whop-Signature` header, `hmac.Equal` for constant-time comparison — mirrors the existing Razorpay/Paddle client pattern exactly.
- `HandleWhopWebhook` now verifies before unmarshaling/processing. **HTTP response is unchanged** — still always `200` (deliberate existing design to avoid Whop retry storms); verification failures are logged and the request returns early before `HandleWebhookEvent` is reached, rather than changing the status code.
- Threaded the new field through the connection metadata read/write paths (`internal/domain/connection/model.go`, `internal/repository/ent/connection.go`) so it actually persists.

## Test plan
- [x] `internal/integration/whop/client_test.go` (new): valid signature passes, invalid signature fails, missing signature header fails, missing configured secret fails with a clear error.
- [x] `internal/api/v1/webhook_whop_test.go` (new): confirms `HandleWhopWebhook` does not reach the processing path when verification fails (missing/invalid signature), and does reach it on a valid signature.
- [x] `go build -mod=mod ./internal/...` clean; `go test -mod=mod ./internal/integration/whop/... ./internal/api/v1/... ./internal/types/...` all green.
- [ ] No mocks needed — `WhopClient` has no generated mock in this repo currently.
- [ ] Manual re-verification against staging recommended: configure a Whop connection with a `webhook_secret`, confirm unsigned requests are now rejected (logged, not processed) while correctly-signed ones still work.

## Note
Whop's actual webhook header name/algorithm (`X-Whop-Signature`, HMAC-SHA256, hex-encoded) was implemented per the standard pattern used by this codebase's other HMAC-based providers (Razorpay) — worth a final check against Whop's live docs before merge in case their exact header name differs.

## Summary by CodeRabbit

* **Security**
  * Whop webhooks are now signature-verified before processing when a webhook secret is configured.
  * When the secret is configured, missing or invalid `X-Whop-Signature` is rejected from further handling (HTTP 200 returned).
  * When no webhook secret is configured, signature verification is skipped and webhooks are allowed.
* **Enhancements**
  * Added encrypted storage/decryption of a Whop webhook secret for verification.
  * Improved connection updates to support partial inputs and Whop webhook-secret-only changes.
* **Tests**
  * Added handler and client test coverage for valid, invalid, missing signatures, and “no secret configured” cases.


agrim123:
## Re: "this flow doesn't touch DB, and we call Whop to validate anyway"

Traced both event handlers in `internal/integration/whop/webhook/handler.go` against this claim — it doesn't hold for either event type this endpoint handles.

### `invoice.paid` → `handleInvoicePaid` (lines 106-152)

No call back to Whop's API anywhere in this path. `data.ID` (attacker-controlled, straight from the forged JSON body) is used directly:
1. `entityIntegrationMappingRepo.List(...)` — DB lookup to resolve a real Flexprice invoice from the Whop invoice ID.
2. `services.InvoiceService.ReconcilePaymentStatus(ctx, flexpriceInvoiceID, types.PaymentStatusSucceeded, &amount)` — **a real DB write** marking that invoice as paid.

If an attacker knows or guesses a real `whop_invoice_id` (not secret — appears in checkout URLs, emails, prior webhook traffic), this marks a real invoice paid with zero payment having occurred. No Whop API round-trip, no DB exclusion — this path is a direct, unauthenticated write to financial state.

### `payment.succeeded` → `handlePaymentSucceeded` (lines 162-220)

This one *does* call Whop: `h.client.GetPlan(ctx, data.Plan.ID)`. But that call only fetches the plan object to read `internal_notes` (which holds the Flexprice `customer_id`) — it does not verify that *this webhook event* is genuine, that a payment actually happened, or that `data.Member.ID` (fully attacker-controlled, never sent to Whop, never checked against anything) is legitimate.

End result: `h.invoiceSyncService.CreateCustomerMapping(ctx, customerID, data.Member.ID)` — **a real DB write** linking a real customer to an arbitrary attacker-supplied `member_id`. Per this file's own doc comment, that mapping is later used so "future invoices for this customer are charged automatically" via the mapped member — i.e. this mapping affects real payment-method resolution for a real customer going forward.

`plan.id` values aren't secret either (visible in Whop checkout URLs), so the only "auth" here is knowing a public-ish identifier, and the Whop API call doesn't function as event verification — it's an unrelated data lookup that happens to use an ID the attacker can supply.

### Conclusion
Both event types reach real DB writes with attacker-controlled data and no cryptographic verification gate. The `GetPlan` call in the `payment.succeeded` path is not equivalent to signature verification — it doesn't check whether the event itself is genuine, only resolves a plan the attacker already knows the ID of. This PR's fix (HMAC-SHA256 signature verification before either handler is reached) closes both paths.

internal/api/v1/webhook.go
```diff
@@ -1218,6 +1218,21 @@ func (h *WebhookHandler) HandleWhopWebhook(c *gin.Context) {
...
+		return
+	}
+
+	signature := c.GetHeader("X-Whop-Signature")
+	if err := whopIntegration.Client.VerifyWebhookSignature(ctx, body, signature); err != nil {
+		h.logger.Error(context.Background(), "Whop webhook signature verification failed",
+			"error", err,
+			"tenant_id", tenantID,
+			"environment_id", environmentID)
+		return
```
subratsahilgupta, line 1234: existing workflow will break for teanants with connection without webhook secret set

subratsahilgupta: Suggestion:

```
// Check if webhook secret is configured in FlexPrice
	hasWebhookSecretConfigured := conn.EncryptedSecretData.Whop != nil &&
		conn.EncryptedSecretData.Whop.WebhookSecret != ""

	// Verify X-Whop-Signature if configured (same optional-secret pattern as Moyasar)
	if hasWebhookSecretConfigured {
		signature := c.GetHeader("X-Whop-Signature")
```

subratsahilgupta (approved):

internal/api/v1/webhook.go
```diff
@@ -1205,36 +1205,79 @@ func (h *WebhookHandler) HandleWhopWebhook(c *gin.Context) {
...
+				"tenant_id", tenantID,
+				"environment_id", environmentID,
+				"note", "signature does not match configured webhook_secret")
+			return
+		}
+
+		h.logger.Info(ctx, "Whop webhook signature verified successfully",
+			"tenant_id", tenantID,
+			"environment_id", environmentID)
+	} else {
```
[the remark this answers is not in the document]
agrim123: This is expected, we are performing migration of those clients.

DiabolusGX (approved):

PR #2270, opened 2026-07-15
fix(clickhouse): batch singleton event inserts

### LINKED ISSUE #2398: ClickHouse creates excessive one-row parts for singleton event deliveries
## Problem

Most `BulkInsertEvents` deliveries contain one event but still use the batch API, producing one-row ClickHouse parts. In production we observed roughly 99,000 one-row parts within ten minutes and CPU throttling during about 70% of scheduling periods.

## Proposed fix

Route singleton bulk deliveries through `AsyncInsert` with `wait=true`, allowing ClickHouse to combine compatible inserts server-side while preserving flush error propagation. Genuine multi-event deliveries keep the existing batch path.

Implementation: #2270

Anri-Lombard:
Fixes #2398

Related to #736

## 📝 Description

You currently route Kafka events through BulkInsertEvents, but most deliveries contain only one event. The batch api still sends a one-row insert.

This creates a lot of background merge pressure, we've seen this result in 99k one-row parts withing ten minutes and CPU throttling during ~70% of scheduling periods.

This PR routes singleton bulk deliveries though ClcikHouse async inserts, then ClickHouse can combine compatible inserts server-side before creating a part.

`wait=true` makes sure the call succeeds after ClickHouse flushed the insert, so errors can propagate. Genuine multi-event deliveries continue using the existing batch implementation.

---

## 🔨 Changes Made
- [ ] Feature 1
- [x] Bug Fix
- [ ] Refactor
- [ ] Documentation Update

---

## ✅ Checklist
- [x] My code follows the project's code style.
- [x] I have added tests where applicable.
- [x] All new and existing tests pass.
- [x] I have updated documentation (if needed).

---

## 📷 Screenshots / Demo (if applicable)
_Add screenshots or demo links here._

---

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized event ingestion for single-event writes by using a faster insert path.
  * Improved overall event write handling for faster, more efficient ingestion.

* **Reliability Improvements**
  * Ensures event inserts wait for server-side batching to flush before reporting completion.

* **Tests**
  * Added coverage to verify singleton bulk inserts use the waiting async insert behavior and correctly propagate insert errors.

### DIFF
--- a/internal/repository/clickhouse/event.go
+++ b/internal/repository/clickhouse/event.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
 	"github.com/flexprice/flexprice/internal/clickhouse"
 	"github.com/flexprice/flexprice/internal/domain/events"
 	ierr "github.com/flexprice/flexprice/internal/errors"
@@ -20,7 +21,7 @@ import (
 )
 
 type EventRepository struct {
-	store  *clickhouse.ClickHouseStore
+	store  interface{ GetConn() driver.Conn }
 	logger *logger.Logger
 }
 
@@ -97,7 +98,8 @@ func (r *EventRepository) InsertEvent(ctx context.Context, event *events.Event) 
 		)
 	`
 
-	err = r.store.GetConn().Exec(ctx, query,
+	// Wait for the server-side batch to flush before acknowledging the insert.
+	err = r.store.GetConn().AsyncInsert(ctx, query, true,
 		event.ID,
 		event.ExternalCustomerID,
 		event.CustomerID,
@@ -128,6 +130,9 @@ func (r *EventRepository) InsertEvent(ctx context.Context, event *events.Event) 
 func (r *EventRepository) BulkInsertEvents(ctx context.Context, events []*events.Event) error {
 	if len(events) == 0 {
 		return nil
+	}
+	if len(events) == 1 {
+		return r.InsertEvent(ctx, events[0])
 	}
 
 	// Start a span for this repository operation
--- a/internal/repository/clickhouse/event_test.go
+++ b/internal/repository/clickhouse/event_test.go
@@ -1,0 +1,67 @@
+package clickhouse
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+	"github.com/flexprice/flexprice/internal/domain/events"
+	"github.com/stretchr/testify/require"
+)
+
+type eventTestStore struct {
+	conn driver.Conn
+}
+
+func (s *eventTestStore) GetConn() driver.Conn {
+	return s.conn
+}
+
+type eventTestConn struct {
+	driver.Conn
+	asyncInsertCalls  int
+	prepareBatchCalls int
+	wait              bool
+	err               error
+}
+
+func (c *eventTestConn) AsyncInsert(_ context.Context, _ string, wait bool, _ ...any) error {
+	c.asyncInsertCalls++
+	c.wait = wait
+	return c.err
+}
+
+func (c *eventTestConn) PrepareBatch(context.Context, string, ...driver.PrepareBatchOption) (driver.Batch, error) {
+	c.prepareBatchCalls++
+	return nil, errors.New("unexpected batch insert")
+}
+
+func TestBulkInsertEventsUsesWaitingAsyncInsertForSingleton(t *testing.T) {
+	insertErr := errors.New("flush failed")
+	for _, test := range []struct {
+		name string
+		err  error
+	}{
+		{name: "success"},
+		{name: "flush error", err: insertErr},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			conn := &eventTestConn{err: test.err}
+			repo := &EventRepository{store: &eventTestStore{conn: conn}}
+			event := events.NewEvent("api_call", "tenant-1", "customer-1", nil, time.Now(), "event-1", "", "test", "environment-1")
+
+			err := repo.BulkInsertEvents(context.Background(), []*events.Event{event})
+
+			if test.err == nil {
+				require.NoError(t, err)
+			} else {
+				require.ErrorContains(t, err, test.err.Error())
+			}
+			require.Equal(t, 1, conn.asyncInsertCalls)
+			require.True(t, conn.wait)
+			require.Zero(t, conn.prepareBatchCalls)
+		})
+	}
+}


Anri-Lombard:
Let me know if regression tests would be valuable here. I'm keeping it as is for now.

Anri-Lombard:
Hi @DiabolusGX — gentle follow-up now that this is updated for review. We’re seeing roughly 99,000 one-row ClickHouse parts within ten minutes and CPU throttling during about 70% of scheduling periods. The change routes only singleton `BulkInsertEvents` calls through `AsyncInsert(wait=true)`; genuine multi-event batches remain unchanged. I added regression coverage for the waiting async-insert path, batch bypass, and error propagation, and `go test ./internal/repository/clickhouse` passes. Would you be willing to confirm whether this approach fits the ClickHouse ingestion direction?

Anri-Lombard:
@omkar273, could you help route this to the right reviewer? The PR is mergeable, limited to two files, and the related production issue is tracked as FLE-1098. I mainly need a direction check on using `AsyncInsert(wait=true)` for singleton deliveries. If the project prefers another approach, I can revise it.

Anri-Lombard:
@DiabolusGX @omkar273, could one of you confirm whether `AsyncInsert(wait=true)` is the right path for single-event `BulkInsertEvents` calls? Multi-event batches are unchanged, and the tests cover the new branch and its error handling. If another maintainer owns this area, please point me to them.

internal/repository/clickhouse/event.go
```diff
@@ -20,7 +21,7 @@ import (
 )
 
 type EventRepository struct {
-	store  *clickhouse.ClickHouseStore
+	store  interface{ GetConn() driver.Conn }
```
agrim123, line 24: Blocking: this weakens the field type solely to allow injecting a fake in the new test.

This file has 13 `r.store.GetConn()` call sites, and the concrete `*clickhouse.ClickHouseStore` also exposes `GetRawConn()` and `Close()` — any future use of those now needs a type assertion. It also diverges from the sibling repositories in this same package (`costsheet_usage.go:20`, `meter_usage.go:24`) which keep the concrete type.

If we do want this seam, please make it a named interface rather than an inline anonymous one:

```go
type connProvider interface {
	GetConn() driver.Conn
}
```

That said, `AGENTS.md` asks that tests at this layer use a real DB rather than mocking the client, so I'd prefer we not introduce the seam at all.

internal/repository/clickhouse/event.go
```diff
@@ -97,7 +98,8 @@ func (r *EventRepository) InsertEvent(ctx context.Context, event *events.Event)
 		)
 	`
 
-	err = r.store.GetConn().Exec(ctx, query,
+	// Wait for the server-side batch to flush before acknowledging the insert.
+	err = r.store.GetConn().AsyncInsert(ctx, query, true,
```
agrim123, line 102: Blocking: two problems here.

**`wait=true` overrides deployed config.** clickhouse-go's `asyncInsert` sets `wait_for_async_insert=1` per-query when `wait` is true, which overrides the profile's explicit `default/wait_for_async_insert: 0` from `clickhouse.yaml:187`. Every insert through this path now blocks for up to `async_insert_busy_timeout_ms` plus flush time, where today it returns immediately.

**This is the synchronous ingest path.** `InsertEvent` is not only reached from the consumer — it serves the ingest API. So this is a direct p99 regression on the hot endpoint, in the opposite direction from the PR's stated goal.

**The client-side call is also redundant.** `default/async_insert: 1` is already set at the profile level and applies to all inserts including the native-protocol batch path, so switching `Exec` to `AsyncInsert` doesn't change whether the insert is async — only whether we block on it.

If we want errors to propagate synchronously, that's a real discussion about the durability/latency tradeoff, but it should be its own PR rather than a side effect of a performance change.

internal/repository/clickhouse/event.go
```diff
@@ -129,6 +131,9 @@ func (r *EventRepository) BulkInsertEvents(ctx context.Context, events []*events
 	if len(events) == 0 {
 		return nil
 	}
+	if len(events) == 1 {
+		return r.InsertEvent(ctx, events[0])
```
agrim123, line 135: Blocking: this makes `BulkInsertEvents` behave differently based on batch size, in ways callers can't see:

1. **Validation order.** `InsertEvent` marshals properties *before* `Validate()`; the bulk path below validates first. An invalid event with unmarshalable properties returns a different error class depending on whether the batch had 1 or 2 events.
2. **Span name.** This path emits `event.insert` instead of `event.bulk_insert`, which will silently split any dashboard or alert currently grouping on bulk ingest.
3. **Durability semantics.** Combined with the `wait=true` change above, one event acks after a server-side flush while two events go through the sync batch path. Same interface method, different guarantees, decided by batch size.

If singleton batching turns out to be the right fix after all (see my main review re: `async_insert` already being enabled server-side), it should preserve the bulk path's validation order and span naming rather than delegating to a method with different semantics.

internal/repository/clickhouse/event_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+			err := repo.BulkInsertEvents(context.Background(), []*events.Event{event})
+
+			if test.err == nil {
+				require.NoError(t, err)
+			} else {
+				require.ErrorContains(t, err, test.err.Error())
+			}
+			require.Equal(t, 1, conn.asyncInsertCalls)
+			require.True(t, conn.wait)
```
agrim123, line 63: These assertions test the wiring rather than the behavior — `conn.wait` and `prepareBatchCalls` lock in the exact implementation, so a refactor breaks the test with no actual regression.

Separately, `eventTestConn` embeds a nil `driver.Conn`, so any method not explicitly overridden panics instead of failing cleanly. The repo struct here also has no `logger`, so if `InsertEvent` gains a log line later this test nil-panics rather than reporting a useful failure.

agrim123 (changes requested):
Thanks for digging into the part-explosion problem — that symptom is worth chasing. But I think this change is aimed at the wrong layer, and as written it would make ingest latency worse without addressing the cause. Four blocking concerns:

## 1. Async inserts are already enabled server-side

The premise that "the batch api still sends a one-row insert" describes the client wire format, not ClickHouse's part-creation behavior. Our Altinity chart already sets async insert at the profile level, which applies to **every** INSERT including native-protocol `PrepareBatch`/`Send()`:

```yaml
# internal/ee/infrastructure/helm/flexprice/templates/infra/clickhouse.yaml:183-187
default/async_insert: 1
default/async_insert_busy_timeout_ms: {{ .Values.clickhouse.altinity.asyncInsert.busyTimeoutMs }}
default/async_insert_max_data_size: {{ .Values.clickhouse.altinity.asyncInsert.maxDataSize }}
default/async_insert_stale_timeout_ms: {{ .Values.clickhouse.altinity.asyncInsert.staleTimeoutMs }}
default/wait_for_async_insert: 0
```

and `values.yaml:528-532` has `asyncInsert.enabled: true`. So the server should already be coalescing these inserts before creating a part.

If you're genuinely seeing 99k one-row parts in ten minutes, something else is going on and this PR won't fix it. Could you share:

- `SELECT count() FROM system.parts WHERE table = 'events' AND active`
- `SELECT * FROM system.asynchronous_insert_log WHERE table = 'events' ORDER BY event_time DESC LIMIT 20`
- `SHOW SETTINGS LIKE 'async_insert%'` **as the application user** on the affected cluster

If `asynchronous_insert_log` has entries, server-side batching is working and the real cause is elsewhere. If it's empty, the profile isn't reaching that cluster/user — which is a config fix, not a Go change.

## 2. `wait=true` inverts the deployed configuration and regresses ingest latency

`AsyncInsert(ctx, query, true, ...)` sets `wait_for_async_insert=1` per-query (clickhouse-go `conn_async_insert.go`), overriding the profile's explicit `wait_for_async_insert: 0`.

This lands on `InsertEvent`, which is the synchronous ingest API path — not just the consumer. Every single-event insert now blocks for up to `async_insert_busy_timeout_ms` plus flush and part-write time, where today it returns immediately. That's a direct latency hit on the hot ingest endpoint, and it serializes the Kafka consumer on each flush.

The PR justifies this as "so errors can propagate," but non-blocking acks are a deliberate choice encoded in the chart. If we want to change that durability/latency tradeoff, it deserves its own discussion and its own PR rather than riding along with a performance fix — especially since it moves latency in the opposite direction from the PR's stated goal.

## 3. Weakening the `store` field to an anonymous interface, for testability only

Changing `store *clickhouse.ClickHouseStore` to `interface{ GetConn() driver.Conn }` exists solely to inject a fake in the new test. Costs:

- Drops the concrete type across **13** `r.store.GetConn()` call sites in this file; any future use of `GetRawConn()` or `Close()` now needs a type assertion.
- Diverges from sibling repositories in the same package (`costsheet_usage.go:20`, `meter_usage.go:24`) which keep the concrete type.
- It's an inline anonymous interface. If we do need this seam, it should at minimum be a named `type connProvider interface { GetConn() driver.Conn }`.

Also worth noting `AGENTS.md` asks that integration tests use a real DB and avoid mocking the client at this layer.

## 4. Silent behavioral divergence between the singleton and multi-event paths

Routing `len(events) == 1` to `InsertEvent` makes `BulkInsertEvents` behave differently depending on batch size, invisibly to callers:

- **Validation order differs.** `InsertEvent` marshals properties *before* calling `Validate()`; the bulk path validates first. Invalid events now surface a different error class depending on batch size.
- **Span name differs.** `event.insert` vs `event.bulk_insert` — this will silently split any dashboard or alert grouping on bulk ingest.
- **Durability semantics differ.** With this change, one event is async and two events are a sync batch, on the same interface method. A caller gets different ack guarantees based on how many events it happened to send.

## Test comments

`require.True(t, conn.wait)` and `require.Zero(t, conn.prepareBatchCalls)` assert the wiring rather than the behavior, so any refactor breaks the test with no actual regression. Separately, `eventTestConn` embeds a nil `driver.Conn`, so any method not explicitly overridden panics rather than failing cleanly — including `InsertEvent` gaining a log line later, since the test repo has no `logger` set.

## Suggested path forward

If the part explosion reproduces:

1. Confirm async insert is actually live on the affected cluster for the app user (queries above). The chart says on; verify the running cluster agrees.
2. If it's off there, fix the config/deploy — no Go changes needed.
3. If it's on and parts still explode, tune `async_insert_busy_timeout_ms` (50ms is quite aggressive; 200–1000ms coalesces substantially more) and `async_insert_max_data_size`.
4. If most Kafka deliveries really are single-event, batch at the **producer/consumer** before reaching ClickHouse. That reduces round trips as well, which the repo-layer change here does not.

Happy to help dig into the ClickHouse-side numbers if you can share the system table output.

Anri-Lombard:
Thanks for the thorough review. I checked the affected deployment using the same ClickHouse user as the application, and the upstream-chart assumption does not hold there:

- `async_insert = false`
- `wait_for_async_insert = true`
- 57,999 new `events` parts in ten minutes, all with one row
- every sampled insert came from the Kafka consumer pods
- `system.asynchronous_insert_log` is not configured

That confirms the root cause is the custom deployment profile, not the repository insert path. I am closing this rather than changing `BulkInsertEvents` semantics, weakening the store type, or overriding the upstream chart's acknowledgement policy per query.

I will address this in the deployment configuration using ClickHouse's native async-insert profile, preserving synchronous acknowledgement initially and measuring part creation, ClickHouse CPU, Kafka lag, and insert latency before tuning the flush window.

PR #2278, opened 2026-07-15
fix(temporal): retry transient temporal client startup errors

### LINKED ISSUE #2399: Temporal client exits during transient frontend startup unhealthiness
## Problem

Temporal's frontend can accept connections before it reports healthy. During that startup window, the eager SDK connection can fail with `Frontend is not healthy yet`, causing FlexPrice to exit even though Temporal becomes healthy shortly afterward.

## Proposed fix

Retry only transient startup connection failures with bounded exponential backoff. Permanent connection errors should still return immediately.

Implementation: #2278

Anri-Lombard:
## **User description**
Fixes #2399

## 📝 Description

Temporal clients were dialled and health-checked during dependency injection for every deployment mode. A transiently unhealthy Temporal frontend could therefore block API and consumer startup, while the original retry covered only the SDK dial and not the later health check.

I changed client construction to use the Temporal SDK's lazy client. API and consumer modes no longer connect to Temporal during dependency injection. Temporal worker modes perform the bounded readiness retry from their startup lifecycle before schedules or workers start. Permanent errors still fail immediately.

## 🔨 Changes Made
- [ ] Feature 1
- [x] Bug Fix
- [x] Refactor
- [ ] Documentation Update

## ✅ Checklist
- [x] My code follows the project's code style.
- [x] I have added tests where applicable.
- [x] All new and existing focused tests pass.
- [ ] I have updated documentation (not needed).

## 📷 Screenshots / Demo (if applicable)
Not applicable.

---
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Temporal service connection reliability with deferred connection checks, bounded startup timeouts, and exponential backoff for temporary failures.
  * Permanent connection errors now fail immediately without unnecessary retries.
  * Startup failures now cleanly stop the service, and shutdown stops workers before closing the connection.
* **Tests**
  * Added coverage for retry behavior, startup cancellation, rollback after schedule failures, and orderly shutdown.

___

## **CodeAnt-AI Description**
Retry transient Temporal startup failures without blocking dependency setup

### What Changed
- Temporal clients are created without connecting during application setup, so temporary frontend outages no longer block dependency initialization
- Worker startup retries temporary unavailable or deadline-exceeded errors for up to two minutes, while permanent errors fail immediately
- Failed worker startup now cleans up the Temporal service, and shutdown stops workers before closing the client
- Added coverage for lazy client creation, retry behavior, cancellation, cleanup, and shutdown order

### Impact
`✅ Fewer application startup failures during temporary Temporal outages`
`✅ Faster failure for invalid Temporal configuration`
`✅ Cleaner worker shutdown after startup errors`
<details><summary><strong>💡 Usage Guide</strong></summary>

### Checking Your Pull Request
Every time you make a pull request, our system automatically looks through it. We check for security issues, mistakes in how you're setting up your infrastructure, and common code problems. We do this to make sure your changes are solid and won't cause any trouble later.

### Talking to CodeAnt AI
Got a question or need a hand with something in your pull request? You can easily get in touch with CodeAnt AI right here. Just type the following in a comment on your pull request, and replace "Your question here" with whatever you want to ask:
<pre>
<code>@codeant-ai ask: Your question here</code>
</pre>
This lets you have a chat with CodeAnt AI about your pull request, making it easier to understand and improve your code.

#### Example
<pre>
<code>@codeant-ai ask: Can you suggest a safer alternative to storing this secret?</code>
</pre>

### Preserve Org Learnings with CodeAnt
You can record team preferences so CodeAnt AI applies them in future reviews. Reply directly to the specific CodeAnt AI suggestion (in the same thread) and replace "Your feedback here" with your input:
<pre>
<code>@codeant-ai: Your feedback here</code>
</pre>
This helps CodeAnt AI learn and adapt to your team's coding style and standards.

#### Example
<pre>
<code>@codeant-ai: Do not flag unused imports.</code>
</pre>

### Retrigger review
Ask CodeAnt AI to review the PR again, by typing:
<pre>
<code>@codeant-ai: review</code>
</pre>

### Check Your Repository Health
To analyze the health of your code repository, visit our dashboard at [https://app.codeant.ai](https://app.codeant.ai). This tool helps you identify potential issues and areas for improvement in your codebase, ensuring your repository maintains high standards of code health.

</details>


Anri-Lombard:
Hi @nkmishra1997 — gentle follow-up on this startup reliability fix. It retries only transient Temporal frontend-not-healthy errors with bounded backoff; permanent connection errors still return immediately. The PR is now rebased onto `develop`, linked to #2399, and `go test ./internal/temporal/client` passes. Would you be willing to take a look and confirm the retry classification and limits?

Anri-Lombard:
Rebased this onto the current `develop`. The PR is now one commit, two files, and mergeable. `go test ./internal/temporal/client` passes. @nkmishra1997, could you take a direction check or route it to the right reviewer?

Anri-Lombard:
@nkmishra1997, could you confirm where the startup retry should live? It currently wraps the SDK dial, but `temporalClient.Start` runs the health check later. Should the retry cover both steps or only the dial? I will adjust the patch once that boundary is clear.

agrim123:
Traced this through Temporal SDK v1.32.1 and verified the behaviour rather than the description. Logic is correct — `client.Dial(o)` is literally `DialContext(context.Background(), o)`, so the swap is behaviour-preserving and just adds deadline support.

Confirmed the classifier works on the real error shape. The production string comes from `internal_workflow_client.go:1298` — `fmt.Errorf("failed reaching server: %w", err)` — and `%w` keeps the chain intact, so `errors.As` reaches through. I probed `isRetryableTemporalDialError` directly:

- wrapped `serviceerror.Unavailable` ("Frontend is not healthy yet") -> `true`
- wrapped raw gRPC `Unavailable` -> `true`
- `NamespaceNotFound` -> `false`
- plain error -> `status.Code` = `Unknown` -> `false` (no accidental retry-everything)

Also dialled a closed port with a 6s budget: 15 attempts, then the error surfaced. The retry loop works against a real server, not only the mocks. And no connection leak on retry — on eager-connect failure the SDK calls `client.Close()` before returning (`internal/client.go:971`). Build, tests, gofmt and vet are clean.

One thing to change before merge.

### 2min blocking dial applies to every deployment mode

`temporalDialTimeout = 2 * time.Minute` blocks inside an FX constructor, and `startServer` takes `temporalClient` as a direct parameter — so FX builds it for **all** modes, including `api` and `consumer`, neither of which starts a Temporal worker.

If Temporal is down, an API-only pod now blocks up to two minutes before failing instead of failing fast. That works against the Kubernetes case this PR targets: two minutes can exceed a `startupProbe` / `initialDelaySeconds` budget, turning a fast crash-loop-and-recover into a slow one. Either bound it lower (~30s), or only dial eagerly where a worker actually runs and use `client.NewLazyClient` elsewhere. I'd lean toward the latter — it fixes the startup race without adding a blocking wait to modes that never talk to Temporal.

### Separate, pre-existing — not caused by this PR

`.env.local:123` sets `FLEXPRICE_TEMPORAL_ENABLED="false"`, but there is no such config field. `TemporalConfig` has no `Enabled`, and the only `temporal_enabled` is `otel.metrics.temporal_enabled` (a tracing flag). The var is dead and the dial is unconditional, which is why the blast radius above is wider than it looks — and it contradicts the AGENTS.md note that Temporal is disabled by default locally. Worth its own issue; happy to open one.

### Minor

`TestDialTemporalClientRetriesUnavailableErrors` returns `(nil, nil)` on success, so it only asserts the attempt count and never checks the client is passed through. Returning a non-nil stub would make it meaningful.

Anri-Lombard:
@agrim123, thanks - I reworked this around the startup boundary you identified.

- Client construction now uses `NewLazyClient`, so API and consumer dependency injection no longer blocks on Temporal.
- `provideTemporalService` only initializes the service.
- Temporal worker startup calls `TemporalService.Start` before schedules or workers are registered.
- The bounded retry now covers the health check and retries only Temporal/gRPC `Unavailable` and `DeadlineExceeded`; permanent errors still fail immediately.
- Worker shutdown now stops both workers and the client.

The focused Temporal and server tests pass after rebasing onto current `develop`, including all four transient error classes, permanent failure, cancellation, lazy construction, and shutdown ordering.

I left the ignored `FLEXPRICE_TEMPORAL_ENABLED` variable out of this PR because it is pre-existing and needs a separate configuration decision.

Anri-Lombard:
I rebased this onto the current `develop` and addressed the remaining startup review points.

- Fx now allows three minutes for the two-minute Temporal readiness retry and the remaining startup hooks.
- Failures after Temporal becomes ready now stop workers and close the client before returning the original startup error.
- The lifecycle regression test runs through `make test`, `make test-verbose`, and `make test-coverage`.

Local `make test`, `make lint-ci`, focused `go vet`, and `git diff --check` pass. @agrim123, could you take another look?

agrim123 (approved):
LGTM. Verified the fix locally: lazy client construction keeps API and consumer modes from connecting to Temporal during DI, the readiness retry now wraps the actual CheckHealth call, permanent errors fail fast, and startup rollback plus ordered shutdown behave as described. Ran the new tests (service, client, cmd/server) along with go build and go vet on the branch, all green. No breaking changes: no exported signatures changed, all GetGlobalTemporalService callers are unaffected, and the removed return-nil-on-error path was a latent bug that this change resolves.

One behavior note for the record: with the lazy client, API and consumer modes now dial Temporal on first RPC rather than at boot, so a Temporal outage surfaces at request time instead of startup. That is the intended tradeoff and callers already handle StartWorkflow errors.

We will merge after internal review.

PR #2321, opened 2026-07-20
fix(helm): order config+SA hooks before migration to unblock fresh install

agrim123:
## Problem

On a **fresh `helm install`** of a legacy-config environment (`.Values.env` unset, ConfigMap rendered), the release deadlocks. The migration Job is a `pre-install`/`pre-upgrade` hook (weight `-5`) that:

- mounts the `<release>-config` ConfigMap, and
- runs as the `flexprice` ServiceAccount.

Both are **plain resources**, which Helm creates only *after* all pre-install hooks. So the migration hook starts before they exist and fails with:

```
MountVolume.SetUp failed for volume "config": configmap "<release>-config" not found
Error creating: pods "..." is forbidden: error looking up service account
  flexprice/flexprice: serviceaccount "flexprice" not found
```

We hit this standing up a new region (europe-west9) and had to work around it with manual pre-creation of the ConfigMap + SA and a `--no-hooks` install + a standalone migration Job.

## Fix

Make the **ConfigMap** and the (single-SA branch) **ServiceAccount** `pre-install,pre-upgrade` hooks at weight **`-10`** — ahead of the migration Job at `-5` — with `hook-delete-policy: before-hook-creation` (**not** `hook-succeeded`, so they persist for the long-lived Deployments that also consume them). The ServiceAccount keeps its Workload Identity / IRSA annotations.

`flexprice-secrets` is unaffected (external / ESO-managed, exists before install).

## Verification

`helm template` hook ordering:

```
weight -10  ConfigMap       <release>-config     ← created first
weight -10  ServiceAccount  flexprice            ← created first
weight  -5  Job             <release>-migration  ← runs after (mounts config, uses SA)
```

Fresh install that previously required manual pre-creation + `--no-hooks` now completes with hooks enabled. Only affects the legacy path; `.Values.env` (config-autobind) environments render no ConfigMap and are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Helm installation and upgrade reliability for database migrations by introducing a dedicated migration ServiceAccount.
  * Ensures the migration ServiceAccount is created before migration hook execution and removed after successful completion.
  * Added support for customizing migration ServiceAccount annotations via chart values, without affecting the hook behavior.

### DIFF
--- a/internal/ee/infrastructure/helm/flexprice/templates/app/configmap.yaml
+++ b/internal/ee/infrastructure/helm/flexprice/templates/app/configmap.yaml
@@ -8,17 +8,6 @@ metadata:
   name: {{ include "flexprice.fullname" . }}-config
   labels:
     {{- include "flexprice.labels" . | nindent 4 }}
-  annotations:
-    # Rendered as a pre-install/pre-upgrade hook (weight -10, ahead of the
-    # migration job at -5) so the migration hook — which mounts this ConfigMap —
-    # can start on a fresh install. Without this, `helm install` deadlocks: the
-    # migration hook runs before the ConfigMap (a normal resource) is created and
-    # fails with `configmap "<release>-config" not found`. hook-delete-policy is
-    # before-hook-creation only (NOT hook-succeeded) so the ConfigMap persists for
-    # the long-lived app Deployments that also mount it.
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-weight": "-10"
-    "helm.sh/hook-delete-policy": before-hook-creation
 data:
   config.yaml: |
     server:
--- a/internal/ee/infrastructure/helm/flexprice/templates/app/configmap.yaml
+++ b/internal/ee/infrastructure/helm/flexprice/templates/app/configmap.yaml
@@ -8,6 +8,17 @@ metadata:
   name: {{ include "flexprice.fullname" . }}-config
   labels:
     {{- include "flexprice.labels" . | nindent 4 }}
+  annotations:
+    # Rendered as a pre-install/pre-upgrade hook (weight -10, ahead of the
+    # migration job at -5) so the migration hook — which mounts this ConfigMap —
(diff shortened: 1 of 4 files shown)


agrim123:
Thanks @coderabbitai — both valid, and the SA one bit us live (a failed upgrade left the primary SA orphaned/deleted, so the Deployments couldn't create pods).

Fixed in 9f65fc9: instead of hooking the **primary** ConfigMap/ServiceAccount (which orphans them on uninstall + deletes on failed upgrade), the primary ones are back to normal release resources, and a **dedicated `<release>-migration` ServiceAccount** is rendered as the pre-install/pre-upgrade hook (weight -10, `hook-delete-policy: before-hook-creation,hook-succeeded` so it self-cleans). It carries the app SA's Workload-Identity annotations, and the migration Job now runs as `<release>-migration`.

On the ConfigMap: left the legacy mount as-is for now — config-autobind (`.Values.env`) environments render no ConfigMap at all, and extracting the ~290-line config body into a shared named template so a `-migration-config` hook can reuse it is a larger refactor better done separately. The migration Job's config dependency only affects legacy first-installs.

agrim123:
Addressed the migration-prereqs findings in ced7b3a:

- **Reserved hook annotations (L24-30):** now rendered *after* the user's `serviceAccount.annotations`, so `helm.sh/hook`, `hook-weight`, and `hook-delete-policy` can't be overridden by user values.
- **WI trust policy (L21-30):** correct — the `<release>-migration` ServiceAccount also needs the cloud IAM trust binding (`system:serviceaccount:<ns>/<release>-migration` → the cloud SA) or the migration hook can't mint tokens (e.g. GMK OAUTHBEARER). Added a note in the commit; this is per-environment infra config (Terraform/IRSA), not chart-renderable. For our euw9 rollout I've already added the `flexprice-migration` KSA to the GSA's `workloadIdentityUser` members.

ojasagarwal-flex (approved):

PR #2389, opened 2026-07-28
feat(events): batch the bulk ingest publish path (producer side)

ojasagarwal-flex:
> **Producer side only.** The bulk consumer is owned by a separate change — see the handoff contract at the bottom.

## Problem

`POST /v1/events/bulk` was only a *transport* convenience. `BulkCreateEvents` looped `CreateEvent`, so a 1000-event request produced **1000 Kafka messages**, and a consumer still had to do one ClickHouse INSERT per event.

On ClickHouse the per-INSERT cost (a part commit plus its metadata round-trips) dominates row cost, so ingest throughput is bounded by *insert count*, not data volume. Measured on a client fleet: ~3,300 events/s of ingest that the pipeline could not absorb, with ClickHouse ~85% idle throughout — the bottleneck was the shape of the writes, not capacity.

## What this changes

**`EventPublisher.PublishBatch`** groups events by partition key, then cuts each group into messages bounded by `max_batch_size` **and** `max_batch_bytes`.

- **Grouping, not blind chunking**, so `Publish()`'s guarantee still holds: a customer's events share a partition.
- **`partition_shards > 1` relaxes that deliberately**, suffixing the key with `hash(event_id) % N`. A tenant with **many** customers already spreads across partitions by key; a tenant with **few** would funnel onto one partition and serialise its batches. The hash (not random) keeps a republished event on the same partition. Safe because handlers are idempotent, must not assume arrival order, and ClickHouse dedupes on event id.
- **The byte bound is not optional**: ~1000 events × ~800 B approaches MSK's 1 MB `max.message.bytes`. An event larger than the cap is still published alone rather than dropped, so it surfaces as a broker rejection instead of silence.
- Publishing stays fire-and-forget to match `CreateEvent`'s 202 contract, but **every event is validated before publish**, so a malformed event fails the HTTP request instead of poisoning a batch that a consumer would retry as a whole.

Also adds `events.EventBatch` (the wire format) and declares `events_bulk` / `events_bulk_dlq` in `kafka.topics` so `migrate kafka` creates them.

Disabled by default (`event_bulk_processing.enabled: false`); when off, `BulkCreateEvents` keeps its existing one-message-per-event behaviour.

## Handoff contract for the consumer change

Three things that are easy to get wrong, one of which I hit while prototyping the consumer:

1. **Do not unmarshal a payload straight into `EventBatch`.** A single-event payload (misroute, replay, or a producer on an older build) decodes into an `EventBatch` with **zero events** — so treating "no events" as nothing-to-do silently acks and drops a billable event. Detect the format first, and decode a single event as a one-event batch. This is noted on the struct doc.
2. **Split the batch on validation, don't fail it.** One malformed event must go to the DLQ individually while the rest of the batch still lands. Returning an error for the whole message retries all 200 events on every attempt and stalls the partition. Batch-level failures can safely fall through to the router's existing retry(3) + poison-queue, because re-inserting is idempotent (ReplacingMergeTree on event id).
3. **`Event.Validate()` requires `id`, `tenant_id`, `event_name` AND `timestamp`.** A bulk event without a timestamp will be rejected — safe only because `ToEvent()` always sets one. Events omitting `tenant_id`/`environment_id` should inherit them from the envelope.

Consumer-side config (`topic_dlq`, `consumer_group`, `rate_limit`) is already declared here so that change needs no config edit. Note **`rate_limit` counts MESSAGES**, each carrying up to `max_batch_size` events — a value tuned for the single-event topic permits 200× more events.

Worth also considering there: `BulkInsertEvents` chunks at 100 rows per INSERT, so a 200-event batch becomes two INSERTs and gives up half the benefit.

## ⚠️ Before enabling in any environment

- **Bulk-ingested events go to `events_bulk`, so they do NOT reach the consumers subscribed to `events` — including meter usage tracking.** Until the bulk consumer (and a batched meter-usage path) exists, enabling this means events are published but not consumed. Do not enable for a deployment that bills from these events.
- **The topics must exist** — declared in `config.yaml`, but on MSK that's a deliberate `migrate kafka` step.

## Testing

7 pure unit tests, no Kafka needed: partition-key grouping, size split, byte split, oversized-single-event still emitted, shard determinism, hot-key spreading across shards, and missing-event-id assignment.

`go build ./...`, `go vet ./internal/...`, `gofmt`, `go test ./internal/kafka ./internal/ee/service ./internal/config`, and `make lint-ci` (loglint merge gate) all clean.

## Summary by CodeRabbit

* **New Features**
  * Added optional bulk event ingestion via a dedicated endpoint, off by default.
  * Introduced Kafka bulk topic/DLQ support plus configurable consumer group, rate limiting, batch sizing, and partition sharding.
  * Bulk requests are published as grouped, size/byte-bounded batches with batch envelopes carrying events plus tenant and environment details.
* **Bug Fixes**
  * Bulk ingestion now only switches to batched publishing when bulk processing is enabled; broker publish failures are logged without failing ingestion.
* **Tests**
  * Added coverage for batching/grouping behavior, topic selection/fallback, missing IDs, sharding/clamping, empty inputs, and environment isolation.

### DIFF
--- a/internal/domain/events/batch.go
+++ b/internal/domain/events/batch.go
@@ -1,0 +1,12 @@
+package events
+
+// EventBatch is the wire format of one `events_bulk` message. Tenant/environment sit on the
+// envelope as a fallback for events that omit them.
+//
+// Consumers must confirm a payload is a batch before decoding: a single event decodes into an
+// EventBatch with zero events, so treating "no events" as nothing-to-do drops a billable event.
+type EventBatch struct {
+	Events        []*Event `json:"events"`
+	TenantID      string   `json:"tenant_id"`
+	EnvironmentID string   `json:"environment_id"`
+}
--- a/internal/kafka/event_publisher.go
+++ b/internal/kafka/event_publisher.go
@@ -40,7 +40,7 @@ func NewEventPublisher(primaryProducer *Producer, secondaryProducer *SecondaryPr
 	}
 	ep := &EventPublisher{
 		logger:     logger,
-		primary:    primaryProducer,   // local cluster is always written
+		primary:    primaryProducer, // local cluster is always written
 		primaryCfg: &cfg.Kafka,
 	}
 	// Presence-based dual-write: a configured second cluster ⇒ also write there.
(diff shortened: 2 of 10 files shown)


ojasagarwal-flex:
### Note: `partition_shards` has no effect yet — leave it at `1`

While sizing this knob I checked how `partition_key` actually reaches Kafka, and it doesn't.

The producer uses `kafka.DefaultMarshaler{}` (`internal/kafka/producer.go:58`), which builds the sarama message with **no `Key`** — `partition_key` travels as a *header* only:

```go
return &sarama.ProducerMessage{
    Topic:   topic,
    Value:   sarama.ByteEncoder(msg.Payload),
    Headers: headers,      // partition_key lives here
}, nil                     // no Key field
```

No custom `Producer.Partitioner` is set anywhere in `internal/kafka`, so sarama's default applies (`partitioner.go:183`):

```go
if message.Key == nil {
    return p.random.Partition(message, numPartitions)   // random, not hashed
}
```

**Consequences**

1. Placement is already random for every message, single or batched. The existing `Publish` comment about a customer's events sharing a partition describes intent, not current behaviour.
2. `partition_shards > 1` cannot spread anything today, and it *does* split one key's events into more groups — smaller batches, which is the opposite of what this PR is for. Default stays `1`.

**Sizing it, if keyed partitioning is enabled later** (switch to `WithPartitioningMarshaler`) — and it should not be `shards = partitions`:

- Shards hash to partitions, so S shards over P partitions reach only `1 − e^(−S/P)` of them: at S = P that's ~63% coverage with 2–3× imbalance. Even spread needs S ≈ 5–10 × P.
- But shards divide the batch: each message carries ~E/S events, so `S ≤ E / max_batch_size` (E = events per key per request). At E = 1000, B = 200 that caps S at 5.
- The batch-size constraint should win, since batching is the point. Practical: start at 4–8, never exceed `E / max_batch_size`, and keep `1` for deployments whose tenants have many customers (cardinality already spreads there).

For scale context: at 200 events/message, 3,400 events/s is ~17 messages/s, so the 6 partitions declared for `events_bulk` are ample — partitions stop being the constraint once batching lands.

Enabling keyed partitioning is a **separate decision**: it would also change placement on the existing `events` topic from random to keyed, redistributing load and potentially concentrating a hot customer onto one partition. Worth measuring on nane2 before it goes anywhere near a client.

internal/publisher/event_publisher.go
```diff
@@ -107,3 +110,44 @@ func (s *eventPublisher) Publish(ctx context.Context, event *events.Event) error
...
+		return nil
+	default:
+		return fmt.Errorf("unknown publish destination: %s", s.config.PublishDestination)
+	}
+}
+
+func (s *eventPublisher) publishEachToDynamo(ctx context.Context, evts []*events.Event) error {
+	var firstErr error
+	for _, e := range evts {
+		if err := s.dynamoPublisher.Publish(ctx, e); err != nil && firstErr == nil {
```
DiabolusGX, line 148: dynamo also has batch writes

internal/kafka/event_publisher.go
```diff
@@ -113,10 +114,196 @@ func (p *EventPublisher) buildMessage(event *events.Event, payload []byte, parti
...
+// groupForBatching buckets events by partition key and cuts each bucket into size- and
+// byte-bounded batches, measuring bytes as each event marshals.
+func groupForBatching(evts []*events.Event, kc *config.KafkaConfig) ([]eventGroup, error) {
+	maxSize := kc.BulkMaxBatchSize
+	if maxSize <= 0 {
+		maxSize = 1
+	}
+	maxBytes := kc.BulkMaxBatchBytes
+
+	order := make([]string, 0)
```
DiabolusGX: order is not required

internal/kafka/event_publisher.go
```diff
@@ -113,10 +114,196 @@ func (p *EventPublisher) buildMessage(event *events.Event, payload []byte, parti
...
+		_, _ = h.Write([]byte(e.ID))
+		key = fmt.Sprintf("%s:s%d", key, h.Sum32()%uint32(shards))
+	}
+	return key
+}
+
+// publishBatchMessage writes one batch to every cluster, resolving the topic per cluster as
+// Publish does — topic names need not match across clusters.
+func (p *EventPublisher) publishBatchMessage(ctx context.Context, payload []byte, g eventGroup, count int) error {
+	build := func() *message.Message {
```
DiabolusGX: build once for primary & secondary

internal/kafka/event_publisher.go
ojasagarwal-flex: Removed. It only existed to make the final flush of still-open groups deterministic (Go randomises map iteration), and nothing depends on that: within a key ordering is already preserved by the mid-stream appends, and across keys the messages are independent. Now iterating the map directly.

internal/kafka/event_publisher.go
ojasagarwal-flex: Done — and this caught a real bug, thanks. My closure called `message.NewMessage(watermill.NewUUID(), …)` **inside** it, so primary and secondary were getting **different message UUIDs**. `Publish` deliberately gives both clusters the same UUID (`event.ID`) so the streams stay dedup-identical; my version broke that invariant while looking like it followed the pattern.

Worth noting for consistency: `Publish` still builds per cluster, with the comment *"the watermill publisher consumes the message it is handed"*. That is defensive rather than load-bearing — `DefaultMarshaler.Marshal` only reads UUID/Metadata/Payload and the sarama publisher never touches the ack channels — so if build-once is right here it is right there too. Happy to apply it to `Publish` and drop the stale comment in a follow-up, rather than leave two conventions in one file.

internal/publisher/event_publisher.go
ojasagarwal-flex: Correct, and my comment was misleading — fixed. DynamoDB does support `BatchWriteItem` (25 items/call); what is missing is on our side: `internal/dynamodb.EventPublisher` exposes only `Publish(ctx, event)`. The comment now says that rather than implying the service has no batch API.

Leaving the actual wiring out of this PR as discussed — it needs a batch method on that client plus its own partial-failure handling (`BatchWriteItem` returns `UnprocessedItems` that must be retried with backoff), which is a change of its own. `publish_destination` is `kafka` in the deployments this targets, so the per-event expansion is not on a hot path today.

internal/domain/events/batch.go
```diff
@@ -0,0 +1,12 @@
+package events
+
+// EventBatch is the wire format of one `events_bulk` message. Tenant/environment sit on the
+// envelope as a fallback for events that omit them.
+//
+// Consumers must confirm a payload is a batch before decoding: a single event decodes into an
+// EventBatch with zero events, so treating "no events" as nothing-to-do drops a billable event.
+type EventBatch struct {
+	Events        []*Event `json:"events"`
```
DiabolusGX, line 9:
```suggestion
	Events        []*Event `json:"data"`
```

DiabolusGX (approved):

PR #2462, opened 2026-08-05
fix(task-import): close SSRF via caller-supplied file_url

omkar273:
## Summary

Closes SFX-0203-F05 / VAPT 6.2 (HIGH) — SSRF via the caller-supplied `file_url` on task import. All four HTTP clients that fetch a task's `file_url` (`FileProcessor.Client`, its two ad-hoc streaming/size-check clients, and `StreamingProcessor.RetryClient`) now share a hardened transport that blocks outbound connections to non-public IPs.

- **`internal/httpclient/ssrf.go`**: `SSRFSafeTransport()` installs a `net.Dialer.Control` hook that validates the actual resolved IP right before `connect()` — for the initial request, every redirect hop, and every candidate address a hostname resolves to — so it can't be bypassed by a check-then-connect DNS-rebinding race. Blocks loopback, link-local (incl. `169.254.169.254` cloud metadata), private (RFC1918/RFC4193), unspecified, and multicast addresses.
- **`internal/httpclient/client.go`**: `ClientConfig` gains an optional `Transport` override (additive, no behavior change for existing callers).
- **`internal/ee/service/file_provider.go`**: restricts `file_url` to `http`/`https` schemes, and fixes `FileProviderRegistry.GetProvider` to match on the parsed hostname instead of substring-matching the raw URL (closes the `evil.com/?x=drive.google.com` misclassification the finding called out).
- **`internal/ee/service/streaming_processor.go` / `file_processor.go` / `task.go`**: wire the hardened transport through all four download paths; `NewFileProcessor`/`NewStreamingProcessor` now build it internally instead of taking an external client.

Scope: only the file-download path is hardened. The shared `httpclient.Client` used by Stripe/HubSpot/Nomod/Whop/Azure Marketplace/Gemini integrations is untouched, since those hit fixed, known-safe hosts.

Design spec and implementation plan (not committed — gitignored under `docs/superpowers/`) walked through the same reasoning in more depth if useful for review context.

## Test plan

- [x] New table-driven tests for `IsPublicIP` (loopback, `169.254.169.254`, RFC1918 x3, IPv6 loopback/ULA, unspecified, multicast, public)
- [x] `httptest.Server`-backed test proving `SSRFSafeTransport()` refuses to connect to a loopback test server
- [x] End-to-end tests proving `FileProcessor.DownloadFile`/`DownloadFileStream`/`GetFileSize` and `StreamingProcessor.downloadFileStream` (the actual production path via `ProcessTaskWithStreaming`) all reject a loopback `file_url`
- [x] Table-driven tests for scheme validation and corrected provider hostname matching (incl. spoofing cases)
- [x] Full `internal/ee/service` suite passes, including `TestTaskService`, confirming the existing mock-client test wiring is unaffected
- [x] `go build`, `go vet` clean; one pre-existing, unrelated `TestWalletService` failure confirmed present on `develop` before this branch (context-cancellation test, untouched by this change)

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened protection against unsafe network requests, including loopback, private, reserved, and other non-public addresses.
  * Reduced risks from proxy settings and hostname-based network bypasses.
  * Restricted file providers to valid HTTP and HTTPS URLs.
  * Improved provider detection to prevent misleading or unrelated URLs from being misidentified.
  * Preserved request tracing when using custom network transports.
  * Prevented streaming downloads and file-size requests from accessing unsafe local endpoints.

___

## **CodeAnt-AI Description**
Block unsafe task file downloads and validate caller-supplied URLs

### What Changed
- Task imports now allow file downloads only from HTTP and HTTPS URLs with a valid host.
- Downloads, streaming requests, size checks, redirects, retries, and provider fetches reject loopback, private, link-local, reserved, and other non-public network addresses.
- Proxy environment settings can no longer bypass the network-address checks.
- File provider detection now uses the actual hostname, preventing lookalike domains and query parameters from being treated as trusted providers.
- Added coverage for unsafe addresses, URL validation, provider spoofing, proxy bypasses, and all file-download paths.

### Impact
`✅ Blocked access to cloud metadata and private network services`
`✅ Safer task imports from caller-supplied URLs`
`✅ Fewer incorrect provider classifications`
<details><summary><strong>💡 Usage Guide</strong></summary>

### Checking Your Pull Request
Every time you make a pull request, our system automatically looks through it. We check for security issues, mistakes in how you're setting up your infrastructure, and common code problems. We do this to make sure your changes are solid and won't cause any trouble later.

### Talking to CodeAnt AI
Got a question or need a hand with something in your pull request? You can easily get in touch with CodeAnt AI right here. Just type the following in a comment on your pull request, and replace "Your question here" with whatever you want to ask:
<pre>
<code>@codeant-ai ask: Your question here</code>
</pre>
This lets you have a chat with CodeAnt AI about your pull request, making it easier to understand and improve your code.

#### Example
<pre>
<code>@codeant-ai ask: Can you suggest a safer alternative to storing this secret?</code>
</pre>

### Preserve Org Learnings with CodeAnt
You can record team preferences so CodeAnt AI applies them in future reviews. Reply directly to the specific CodeAnt AI suggestion (in the same thread) and replace "Your feedback here" with your input:
<pre>
<code>@codeant-ai: Your feedback here</code>
</pre>
This helps CodeAnt AI learn and adapt to your team's coding style and standards.

#### Example
<pre>
<code>@codeant-ai: Do not flag unused imports.</code>
</pre>

### Retrigger review
Ask CodeAnt AI to review the PR again, by typing:
<pre>
<code>@codeant-ai: review</code>
</pre>

### Check Your Repository Health
To analyze the health of your code repository, visit our dashboard at [https://app.codeant.ai](https://app.codeant.ai). This tool helps you identify potential issues and areas for improvement in your codebase, ensuring your repository maintains high standards of code health.

</details>

### DIFF
--- a/internal/ee/service/file_processor.go
+++ b/internal/ee/service/file_processor.go
@@ -41,18 +41,19 @@ type FileProcessor struct {
 // Default settings:
 // - MaxMemoryFileSize: 10MB (files smaller than this are processed in memory)
 // - MaxFileSize: 1GB (maximum file size allowed)
-func NewFileProcessor(client httpclient.Client, logger *logger.Logger) *FileProcessor {
+func NewFileProcessor(logger *logger.Logger) *FileProcessor {
+	streamingProcessor := NewStreamingProcessor(logger)
+
 	// Configure retryable HTTP client
 	retryClient := retryablehttp.NewClient()
 	retryClient.RetryMax = 3
 	retryClient.RetryWaitMin = 1 * time.Second
 	retryClient.RetryWaitMax = 30 * time.Second
 	retryClient.Logger = logger.GetRetryableHTTPLogger()
-	// Instrument outbound file downloads for SigNoz External API Monitoring.
-	retryClient.HTTPClient.Transport = httpclient.OtelTransport(retryClient.HTTPClient.Transport)
+	retryClient.HTTPClient.Transport = streamingProcessor.Transport
 
 	return &FileProcessor{
-		StreamingProcessor: NewStreamingProcessor(client, logger),
+		StreamingProcessor: streamingProcessor,
 		ProviderRegistry:   NewFileProviderRegistry(),
 		CSVProcessor:       NewCSVProcessor(logger),
 		JSONProcessor:      NewJSONProcessor(logger),
@@ -171,7 +172,7 @@ func (fp *FileProcessor) DownloadFileStream(ctx context.Context, t *task.Task) (
 	// Make the request with extended timeout for large file downloads
 	httpClient := &http.Client{
 		Timeout:   10 * time.Minute, // Extended timeout for large file downloads
-		Transport: httpclient.OtelTransport(nil),
+		Transport: fp.Transport,
 	}
 	resp, err := httpClient.Do(httpReq)
 	if err != nil {
@@ -237,7 +238,7 @@ func (fp *FileProcessor) GetFileSize(ctx context.Context, t *task.Task) (int64, 
 
 	httpClient := &http.Client{
 		Timeout:   30 * time.Second, // Shorter timeout for HEAD requests
-		Transport: httpclient.OtelTransport(nil),
+		Transport: fp.Transport,
 	}
 	resp, err := httpClient.Do(httpReq)
 	if err != nil {
--- a/internal/ee/service/file_processor_test.go
+++ b/internal/ee/service/file_processor_test.go
@@ -1,0 +1,67 @@
+package service
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/flexprice/flexprice/internal/config"
+	"github.com/flexprice/flexprice/internal/domain/task"
+	"github.com/flexprice/flexprice/internal/logger"
+)
+
+func newTestFileProcessor(t *testing.T) *FileProcessor {
+	t.Helper()
+	cfg := config.GetDefaultConfig()
+	log, err := logger.NewLogger(cfg)
+	if err != nil {
+		t.Fatalf("failed to build logger: %v", err)
+	}
+	// DownloadFile/DownloadFileStream/GetFileSize each issue a single
+	// request (fp.Client.Send or a plain http.Client.Do) -- no
+	// retryablehttp.Client is involved, so no retry-backoff tuning is
+	// needed here (contrast with streaming_processor_test.go, which does
+	// exercise RetryClient).
+	return NewFileProcessor(log)
+}
+
+func newLoopbackServer(t *testing.T) *httptest.Server {
+	t.Helper()
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte("a,b\n1,2\n"))
+	}))
+	t.Cleanup(srv.Close)
+	return srv
+}
+
+func TestFileProcessor_DownloadFile_BlocksLoopback(t *testing.T) {
+	srv := newLoopbackServer(t)
+	fp := newTestFileProcessor(t)
+	tsk := &task.Task{FileURL: srv.URL + "/file.csv"}
+
+	if _, err := fp.DownloadFile(context.Background(), tsk); err == nil {
+		t.Fatal("expected DownloadFile to a loopback address to be blocked, got nil error")
+	}
+}
+
+func TestFileProcessor_DownloadFileStream_BlocksLoopback(t *testing.T) {
+	srv := newLoopbackServer(t)
+	fp := newTestFileProcessor(t)
+	tsk := &task.Task{FileURL: srv.URL + "/file.csv"}
+
+	if _, err := fp.DownloadFileStream(context.Background(), tsk); err == nil {
+		t.Fatal("expected DownloadFileStream to a loopback address to be blocked, got nil error")
+	}
+}
+
+func TestFileProcessor_GetFileSize_BlocksLoopback(t *testing.T) {
+	srv := newLoopbackServer(t)
+	fp := newTestFileProcessor(t)
+	tsk := &task.Task{FileURL: srv.URL + "/file.csv"}
+
+	if _, err := fp.GetFileSize(context.Background(), tsk); err == nil {
+		t.Fatal("expected GetFileSize to a loopback address to be blocked, got nil error")
+	}
+}
--- a/internal/ee/service/file_processor_test.go
+++ b/internal/ee/service/file_processor_test.go
@@ -18,11 +18,6 @@ func newTestFileProcessor(t *testing.T) *FileProcessor {
 	if err != nil {
 		t.Fatalf("failed to build logger: %v", err)
 	}
-	// DownloadFile/DownloadFileStream/GetFileSize each issue a single
-	// request (fp.Client.Send or a plain http.Client.Do) -- no
-	// retryablehttp.Client is involved, so no retry-backoff tuning is
-	// needed here (contrast with streaming_processor_test.go, which does
-	// exercise RetryClient).
 	return NewFileProcessor(log)
 }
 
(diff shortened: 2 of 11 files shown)


omkar273:
Addressed the automated review findings:

- **Proxy bypass** (DiffSniff, CodeAnt — duplicate): fixed in 16b4f990b. `SSRFSafeTransport()` now sets `transport.Proxy = nil` so an `HTTP_PROXY`/`HTTPS_PROXY` env var can't make the dial guard validate the proxy's IP while the proxy forwards to an internal target.
- **`IsPublicIP` range gaps** (CodeAnt, CodeRabbit — duplicate): fixed in 2b41921fc. Now also rejects RFC6598 shared address space (`100.64.0.0/10`), RFC5737/RFC3849 documentation ranges, RFC2544 benchmarking, and the limited broadcast address.
- **S3 host matching too broad** (CodeRabbit): fixed in 2b41921fc. No longer matches other `*.amazonaws.com` services (ec2, lambda, ...) or unrelated domains containing `s3.`.
- **Hostless URLs accepted** (CodeRabbit): fixed in 2b41921fc. `parseAllowedURL` now rejects `http://`/`http:opaque`-style URLs with no host.

Not applying the **"IP disclosure in error messages"** suggestion (Medium): verified against `internal/rest/middleware/errhandler.go` — `getDisplayMessage` only ever returns a `WithHint(...)` string or a static fallback, never the raw wrapped error, and every download path here sets a fixed hint plus `WithReportableDetails({"provider": ...})` only. The guard's error text (with the resolved IP) never reaches the HTTP response; it only shows up in server-side structured logs, which is the correct place for it — that's what lets operators investigate a blocked SSRF attempt. Redacting it there would remove real diagnostic value for no actual external exposure.

All new/updated tests pass; full `internal/httpclient` + `internal/ee/service` suite is green aside from one pre-existing, unrelated `TestWalletService` failure (confirmed present on `develop` before this branch).

omkar273:
Ran an independent senior-engineer review pass (separate from the automated bots) after the fixes above. Two more real gaps in the same vein, fixed in 5720505f7:

- `IsPublicIP` didn't unwrap NAT64 (`64:ff9b::/96`, RFC6052) or 6to4 (`2002::/16`, RFC3056) IPv6 encodings of an IPv4 address — e.g. `64:ff9b::a9fe:a9fe` (NAT64 encoding of `169.254.169.254`) and `2002:7f00:1::` (6to4 encoding of `127.0.0.1`) both classified as public. Fixed by unwrapping the embedded IPv4 address and re-validating it recursively, rather than blocking the whole range outright (which would break legitimate egress in IPv6-only/NAT64 deployments).
- `0.0.0.0/8` ("this network" — only the exact `0.0.0.0` was caught before) and `240.0.0.0/4` (Class E reserved) weren't blocked.

Also confirmed by that review, with no code changes needed:
- The four download call sites (`FileProcessor.Client`, its two ad-hoc clients, `StreamingProcessor.RetryClient`) all resolve to the exact same `SSRFSafeTransport()` instance — traced end to end, no unguarded transport sneaks in anywhere.
- HTTP/2 connection coalescing, CONNECT tunneling, and mixed public/private DNS answers are all dead ends given this architecture (each candidate IP is independently validated by `Control` at actual dial time).
- The `ErrorSummary` set on a blocked download (which does include the resolved IP, e.g. `"ssrf guard: connections to 127.0.0.1 are not allowed"`) is never persisted or returned via the API on the failure path — confirms the earlier-rejected "IP disclosure" bot finding really doesn't hold up.

Two things came up that are real but out of scope for an SSRF fix specifically, so I've filed them as separate follow-ups rather than growing this PR further:
- File downloads have no response-size cap (a public-but-malicious server could return an unbounded response) — `MaxFileSize` exists but isn't wired into the actual byte stream anywhere.
- `FileProcessor.DownloadFile`/`DownloadFileStream`/`GetFileSize` and its own `RetryClient` field are dead code — production task processing only ever goes through `StreamingProcessor.downloadFileStream`. Worth deciding whether to wire the size-based dispatch in as the doc comment implies, or delete the duplication.

All tests green (`internal/httpclient` + `internal/ee/service`), same one pre-existing unrelated `TestWalletService` failure as before.

agrim123 (commented):
Reviewed adversarially — focused on trying to get *past* the guard. **Verdict: this substantively closes VAPT 6.2 (SSRF via `file_url`). LGTM.** The dial-time `net.Dialer.Control` architecture is the right call and I couldn't find an IP-guard bypass.

### What holds up

- **Dial-time validation is the correct design.** `controlBlockNonPublic` as `net.Dialer.Control` fires per-connection after DNS resolution, before `connect()`, for the initial request **and every redirect hop and every candidate IP** a name resolves to. That closes the check-then-connect DNS-rebinding TOCTOU by construction — there's no window because the check *is* at connect time on the resolved address.
- **All four fetch sites are guarded** — `DownloadFile` (`Client` now built from `SSRFSafeTransport`), `DownloadFileStream`, `GetFileSize` (both via `fp.Transport`), and `StreamingProcessor.RetryClient`. Removing the external-client injection so callers can't pass an unguarded client is the right move — no unguarded download path left. `/tasks/:id/download` is `s3://`-only presigned, correctly unaffected.
- **`IsPublicIP` — no bypass found.** I ran the exact logic against ~20 crafted inputs. The subtle ones are all handled:
  - `::ffff:127.0.0.1` / `::ffff:169.254.169.254` (IPv4-mapped IPv6) → blocked, because `To4()` unwraps and `IsLoopback`/`IsPrivate` inspect the embedded v4. This was my sharpest concern and it's correct.
  - NAT64 `64:ff9b::a9fe:a9fe` and 6to4 `2002:…` encodings of internal IPs → blocked via `embeddedV4` + recursion. Nice catch including these.
  - 169.254.169.254 IMDS, RFC1918/4193, CGNAT 100.64/10, TEST-NETs, class E, multicast, unspecified → all blocked.
- **`transport.Proxy = nil`** is necessary (a set `HTTP(S)_PROXY` would validate the proxy IP while it forwards to the internal target) — and there's a regression test for it. Good.
- **Scheme allowlist + hostname-based `GetProvider`** close `file://`/`gopher://`/`ftp://` and the `evil.com/?x=drive.google.com` misclassification.
- Tests are genuinely adversarial (httptest loopback block, proxy-env, scheme/host-spoof tables, mapped/NAT64 metadata), not happy-path.

### Residual gaps — none reopen the SSRF finding, all fine as follow-ups

1. **Unbounded `io.ReadAll` in `internal/httpclient/client.go:103`** (pre-existing, out of scope for 6.2). The in-memory `DownloadFile` path reads the whole body with no cap, so a *public* URL (which the guard lets through) returning a huge/infinite stream can OOM the worker. `DownloadFileStream`/`GetFileSize` enforce `MaxFileSize`; this path doesn't. Separate DoS class — worth a follow-up ticket, not a blocker here.
2. **No explicit redirect-count cap** — relies on Go's default 10. Each hop is re-validated by the Control hook so it's not an SSRF hole; an explicit `CheckRedirect` cap would be belt-and-suspenders.
3. **No explicit redirect-to-internal integration test** — the per-hop guard makes this structurally safe, but a test that 302s to `127.0.0.1` would be stronger evidence.

Didn't build/run locally (read-only review). None of the above blocks merge for 6.2.

omkar273:
Addressed the remaining two items:

**NAT64 `64:ff9b:1::/48` not caught** (CodeRabbit): confirmed — this is RFC 8215's "Local-Use IPv4/IPv6 Translation Prefix," a real, standardized alternate NAT64 prefix distinct from the RFC 6052 well-known `64:ff9b::/96` one. Rather than generalizing to arbitrary organization-specific NSPs at other prefix lengths (32/40/48/56/64-bit, each with a different RFC 6052 embedding format around the reserved "u" byte — unbounded complexity for an attack surface that requires the attacker to already know a deployment-private prefix), this one specific RFC-registered prefix is blocked outright rather than unwrapped, since decoding its split embedding is unnecessary complexity for a prefix that isn't globally routable regardless of payload. Added test cases for both a metadata-embedding and a "looks public" address under this prefix — both correctly blocked.

**Correction to my earlier reply on the "IP disclosure" finding**: I said I wasn't applying that suggestion, reasoning that the guard's error text never reaches the external API response. That's still true, but on reflection the redaction is worth doing anyway as defense-in-depth for the internal logs/traces themselves — and it turns out to cost nothing: every call site that logs a download failure already logs `"url", downloadURL` as a separate structured field, so operators keep full diagnostic visibility regardless of whether the guard's own error string also repeats it. `controlBlockNonPublic`'s errors no longer include the resolved address/IP; added a test asserting the error text never contains it.

All 9 inline review comments are now addressed. Full `internal/httpclient` + `internal/ee/service` suite green, same one pre-existing unrelated `TestWalletService` failure as before, `go build`/`go vet`/`gofmt` clean.

PR #2494, opened 2026-08-09
fix(authz): remove X-Environment-ID fallback + guard environment list routes

agrim123:
## Summary
Fixes two authorization findings identified during internal security testing:

- **Environment ID header not validated**: `internal/rest/middleware/auth.go` accepted a caller-supplied `X-Environment-ID` header verbatim whenever a credential (JWT user, config-map API key) had no bound environment ID. Confirmed no legitimate SDK/frontend caller relies on this header — removed outright.
- **Missing permission check on environment listing**: `GET /v1/environments` and `GET /v1/environments/:id` had no RBAC permission check, unlike every write route on the same resource. Now gated behind `environment:read`, same pattern as the existing `write` checks.

**Scope note**: the second fix closes the exploit path for DB-backed service-account keys. JWT users and config-map API keys bypass RBAC entirely by design in this codebase (`permission.go:41-42`) — that is a separate, already-tracked issue and is not addressed by this PR.

Related ClickHouse SQL injection findings were fixed separately in #2253 (already merged to `develop`; this branch is rebased on top of it).

## Out of scope (follow-ups to file)
- Full audit of every unguarded GET route in `router.go` (broader than these two findings).
- `UserEnvMapping` fail-open default (dead config today, not itself exploitable once the header fallback above is gone).
- JWT/config-key RBAC bypass (empty roles) — separate known issue, still open.
- A handful of dev-only docs still instruct engineers to pass `X-Environment-ID` alongside a JWT/config key — now silently inert, should be corrected.
- The router-level wiring itself (the two `read(...)` guard placements) has no end-to-end regression test — the new tests exercise `RequirePermission` in isolation via the existing test helper, consistent with prior convention in this file.

## Test plan
- [x] `internal/rest/middleware/auth_test.go` — subtest confirms `X-Environment-ID` header has no effect when the JWT claim is absent.
- [x] `internal/rest/middleware/permission_test.go` — new tests confirm a minimal service-account role is denied and `super_admin` is allowed on `environment:read`.
- [x] `go test ./internal/... -race` green (56 packages, 0 failures).
- [x] `go build`, `go vet`, `gofmt`, `make lint-ci` all clean.

### DIFF
--- a/internal/api/router.go
+++ b/internal/api/router.go
@@ -112,6 +112,7 @@ func NewRouter(
 	// Initialize permission middleware
 	permissionMW := middleware.NewPermissionMiddleware(rbacService, logger)
 	write := permissionMW.RequirePermission // shorthand used on every write route
+	read := permissionMW.RequirePermission  // shorthand used on read-guarded routes
 
 	// Add middleware to set swagger host dynamically
 	router.Use(func(c *gin.Context) {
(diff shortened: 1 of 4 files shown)


internal/rest/middleware/auth.go
```diff
@@ -52,11 +52,6 @@ func setContextValues(c *gin.Context, tenantID, userID, environmentID, userType
 		ctx = context.WithValue(ctx, types.CtxUserType, userType)
 	}
 
-	// Set additional headers for downstream handlers
-	if environmentID == "" {
-		environmentID = c.GetHeader(types.HeaderEnvironment)
```
DiabolusGX, line 57: I think we need this. JWT does not have environment id, there's only 1 auth for all environments for user account, but in case of api-key based auth, we can skip this fallback.

internal/api/router.go
```diff
@@ -112,6 +112,7 @@ func NewRouter(
 	// Initialize permission middleware
 	permissionMW := middleware.NewPermissionMiddleware(rbacService, logger)
 	write := permissionMW.RequirePermission // shorthand used on every write route
+	read := permissionMW.RequirePermission  // shorthand used on read-guarded routes
```
DiabolusGX, line 115: then we should just call it `perms` instead of `read` and `write`

PR #2533, opened 2026-08-12
fix(helm): repair topology spread selector and kafka broker list

agrim123:
Two defects in the Helm chart's rendered config. Both were invisible in practice because something else was compensating — which is what makes them worth fixing now rather than when the compensation goes away.

## 1. Topology spread constraints matched zero pods

`values.yaml` hardcoded the selector:

```yaml
topologySpreadConstraints:
  - maxSkew: 1
    topologyKey: topology.kubernetes.io/zone
    whenUnsatisfiable: ScheduleAnyway
    labelSelector:
      matchLabels:
        app.kubernetes.io/name: flexprice   # <-- hardcoded
```

Pods are labelled via `flexprice.name`, which honours `nameOverride`. Any deployment setting `nameOverride` therefore got a selector matching **no pods at all**.

Because the default is `whenUnsatisfiable: ScheduleAnyway`, the scheduler silently ignores an unsatisfiable constraint — no Pending pods, no events, nothing in the manifest to indicate spreading was lost. Every replica could land in a single AZ. It "worked" only for deployments that left `nameOverride` unset, where the chart name happens to equal the label.

**Fix:** drop the selector from `values.yaml`; inject it per component via a new `flexprice.topologySpreadConstraints` helper — selector labels (honouring `nameOverride`) plus the component key, so api/consumer/worker/frontend each spread against their own replicas instead of being balanced as one pool.

Preserved behaviour:
- a constraint defining its own `labelSelector` passes through untouched
- `<component>.topologySpreadConstraints` still overrides the global list
- an empty list renders no block

## 2. `kafka.brokers` rendered as one malformed address

```yaml
brokers:
  - {{ include "flexprice.kafkaBrokers" . | quote }}
```

`flexprice.kafkaBrokers` returns the comma-joined form that `FLEXPRICE_KAFKA_BROKERS` needs — correct for the env var, wrong for the config file. As a single list item it yields one entry, `"b1:9093,b2:9093,b3:9093"`, which sarama treats as a single hostname.

This never surfaced because the env var outranks the config file at `Unmarshal` time and viper's default `StringToSliceHookFunc(",")` splits it correctly. I confirmed the precedence directly against viper: from the config file alone the value stays a single malformed broker; with the env var set it splits into three. So the env var is the only reason multi-broker clusters work today — remove or rename it and the cluster silently becomes unreachable.

**Fix:** emit one YAML entry per broker.

## Verification

`helm lint` clean. `helm template` checks:

- selector tracks `nameOverride` — the previously broken case
- each component gets its own `app.kubernetes.io/component` in the selector
- operator-supplied `labelSelector` passes through unmodified
- per-component override wins; other components keep the global default
- `topologySpreadConstraints: []` renders no block
- rendered `config.yaml` parses and yields three distinct brokers

No values-schema change: `topologySpreadConstraints` keeps its shape, only the redundant hardcoded selector is gone. Existing values files that set their own `labelSelector` are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kafka broker configuration rendering so multiple brokers are correctly represented as separate YAML entries.
  * Improved Kubernetes pod distribution settings across API, consumer, frontend, and worker deployments.
  * Deployment labels and selectors are now applied consistently, including when custom naming is configured.
  * Explicit topology selector settings are preserved, while invalid configuration formats are safely ignored.

### DIFF
--- a/internal/ee/infrastructure/helm/flexprice/templates/_helpers.tpl
+++ b/internal/ee/infrastructure/helm/flexprice/templates/_helpers.tpl
@@ -772,14 +772,26 @@ could land in one AZ with nothing in the manifest to show for it.
 
 A constraint that DOES define labelSelector is passed through untouched, so an
 operator can still spread against an arbitrary set of pods.
+
+Presence is tested with `hasKey`, not truthiness: an explicit `labelSelector: {}`
+is a VALID selector meaning "match every pod in the namespace", but it is falsey
+in Go templates. Testing truthiness would take the injection branch and emit
+`labelSelector` twice in the same constraint — YAML resolves the duplicate to the
+last occurrence, so the operator's explicit selector would be silently discarded
+rather than rejected.
 */}}
 {{- define "flexprice.topologySpreadConstraints" -}}
 {{- $ctx := .ctx -}}
 {{- $component := .component -}}
 {{- $cfg := index $ctx.Values $component | default dict -}}
 {{- $tsc := default $ctx.Values.topologySpreadConstraints (get $cfg "topologySpreadConstraints") -}}
+{{- /* `--set topologySpreadConstraints=[]` yields the STRING "[]", not an empty
+       list, and `range` over a string is a render error. Anything that is not a
+       real list is treated as "no constraints" so the caller's `with` emits
+       nothing, matching `topologySpreadConstraints: []` set via a values file. */ -}}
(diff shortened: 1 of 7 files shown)


agrim123:
Opened against the wrong remote — closing. Will re-target.

ojasagarwal-flex (approved):

PR #2658, opened 2026-08-25
feat(meter_usage): publish records to analytics lake topic (additive, fire-and-forget)

agrim123:
## What

Publish `meter_usage` records to a Kafka topic (`analytics.meter_usage`) after they're written to ClickHouse, so an analytics lake can ingest them. **Additive, fire-and-forget, dark by default.**

## Why

Feeds an open-format (Iceberg) analytics lake that unifies meter_usage + Postgres dimensions for revenue/usage analytics, shippable as a BYOC feature. This is the app-side hook; the lake infra is separate.

## How

- New `internal/kafka/MeterUsagePublisher` — thin type over the existing `messagePublisher` (reuses the local-cluster producer; the lake is just another topic).
- `MeterUsageRepository.BulkInsertMeterUsage` publishes the same records **after** the CH batch commits.
- **Presence-gated**: `kafka.lake_topic` empty (default) → publisher is `nil` → whole path is a no-op. CH insert runs exactly as before.
- **Fire-and-forget** (correctness boundary, tested): `PublishMeterUsage` never returns a publish error, and the repo discards its return — structurally cannot fail `BulkInsertMeterUsage` or affect billing.
- Constructor made **variadic** → no existing caller breaks; no mock regen needed.
- `ingested_at` stamped in Go at publish time (records rely on the CH DEFAULT) to give the lake a version column.

## Enable

Set `FLEXPRICE_KAFKA_LAKE_TOPIC=analytics.meter_usage`. Unset = dark no-op.

## Tests

`internal/kafka/meter_usage_publisher_test.go` — topic+payload marshal, fire-and-forget error-swallow, nil no-op. All pass.

## Risk

Ships dark; zero behavior change until the env var is set. A lake-publish failure is logged and swallowed, never propagated.

### DIFF
--- a/internal/kafka/meter_usage_publisher.go
+++ b/internal/kafka/meter_usage_publisher.go
@@ -1,0 +1,76 @@
+package kafka
+
+import (
+	"context"
+	"encoding/json"
+	"time"
+
+	"github.com/ThreeDotsLabs/watermill/message"
+	"github.com/flexprice/flexprice/internal/config"
+	"github.com/flexprice/flexprice/internal/domain/events"
+	"github.com/flexprice/flexprice/internal/logger"
+)
+
+// MeterUsagePublisher publishes meter_usage records to the analytics lake topic
+// (cfg.Kafka.LakeTopic) AFTER they are written to ClickHouse. It is ADDITIVE and
+// FIRE-AND-FORGET: ClickHouse stays authoritative, so a publish failure here is logged
+// and swallowed — it must never fail the caller or affect billing. Presence-gated: when the
+// lake topic is not configured the fx provider returns nil, and every method is a no-op.
+type MeterUsagePublisher struct {
+	publisher messagePublisher
+	topic     string
+	logger    *logger.Logger
+}
+
+// NewMeterUsagePublisher is the fx provider. It reuses the local-cluster producer that already
+// carries source events. Presence-based: an empty cfg.Kafka.LakeTopic ⇒ nil publisher (no-op),
+// mirroring the KafkaSecondary dual-write gating.
+func NewMeterUsagePublisher(primaryProducer *Producer, cfg *config.Configuration, logger *logger.Logger) *MeterUsagePublisher {
+	if cfg.Kafka.LakeTopic == "" || primaryProducer == nil {
+		return nil
+	}
+	return &MeterUsagePublisher{
+		publisher: primaryProducer,
+		topic:     cfg.Kafka.LakeTopic,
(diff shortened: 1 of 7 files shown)


internal/kafka/meter_usage_publisher.go
DiabolusGX: Can exist in `meter_usage_tracking` service which will just publish it's own processed message

internal/kafka/meter_usage_publisher.go
```diff
@@ -0,0 +1,76 @@
...
+// event id (the lake's dedup key). ingested_at is stamped when unset so the lake has a version
+// column. Fire-and-forget: per-record errors are logged, never returned.
+func (p *MeterUsagePublisher) PublishMeterUsage(ctx context.Context, records []*events.MeterUsage) error {
+	// nil publisher (lake not configured) ⇒ no-op.
+	if p == nil || p.publisher == nil {
+		return nil
+	}
+
+	for _, record := range records {
+		// The lake dedups on (id, ingested_at). BulkInsert relies on the ClickHouse DEFAULT for
```
DiabolusGX, line 49: we'll dedupe on event id + event timestamp right? not on even ingested at

agrim123: Good catch — and yes, you're right on the identity key. Two separate things here:

**Dedup identity** (which rows are the same event) = `(event_id, event_timestamp)` — matches the CH `meter_usage` `ORDER BY (tenant, env, customer, meter, timestamp, id)`. The lake's `_v` dedup view partitions on `(id, timestamp)`, exactly as you say.

**Version tiebreaker** (which copy of a duplicate wins) = `ingested_at`. This mirrors the CH table's `ReplacingMergeTree(ingested_at)` — `ingested_at` is the `ver` column. `event_timestamp` can't be the tiebreaker because it's immutable (same for every copy of an event), so it gives the merge nothing to order on.

So `ingested_at` here isn't the identity — it's the version, matching CH's `ReplacingMergeTree(ingested_at)` semantics. The lake view keys identity on `(id, timestamp)` and orders by `ingested_at DESC`.

One subtlety worth flagging: this publish stamps `ingested_at` at publish time when it's zero (the CH DEFAULT fills it on insert, so records reach here unset). That keeps it monotonic-per-event for the version ordering. If we ever need the lake's `ingested_at` to be byte-identical to CH's, we'd read it back post-insert instead — but for latest-wins dedup, publish-time is sufficient and correct.

DiabolusGX (approved):

PR #2659, opened 2026-08-26
feat(reprocess): enforce event_timestamp immutability per event_id (R1 guard)

agrim123:
## What

Enforce **`event_timestamp` immutability per `event_id`** in the raw-events reprocess path: a republish must not change the business timestamp of an existing event.

## Why (R1)

A live dedup test on the analytics lake proved a real double-count: if an event is republished with the same `event_id` but a **changed** `event_timestamp`, the lake's Iceberg dedup keys on `(id, timestamp)`, so the changed timestamp becomes a **new identity** — both copies survive and `sum(qty)` over-counts. No lake-side fix is possible; the guard must be producer-side. First-time ingest can't violate this (nothing to mutate) — the only place a republish happens is reprocessing.

## How

In `ReprocessRawEvents`, right before re-publishing each transformed event:

```go
if transformedEvent.ID != "" && !transformedEvent.Timestamp.Equal(rawEvent.Timestamp) {
    // R1 violation: skip publish, count it, log ERROR
    continue
}
```

- `.Equal()` (not `==`) — monotonic/location-safe; fires only on a genuine instant difference.
- Mirrors the existing per-event skip idiom (`continue` + counter) — **does not fail the batch**.
- Adds `TotalR1Violations` to `ReprocessRawEventsResult`, surfaced in batch/final summary logs and the Temporal workflow result.
- **No DB lookup** — both timestamps are already in hand (original `rawEvent` + candidate `transformedEvent`). Off the hot path.

Normal reprocessing (timestamp unchanged) is a **no-op** — the transformer derives `Timestamp` from the same source, so `.Equal()` is true and it publishes exactly as before.

## Tests

`raw_events_reprocessing_test.go` (TDD):
- `TestReprocess_TimestampUnchanged_Publishes` — healthy → published, `r1_violations=0`.
- `TestReprocess_TimestampChanged_SkippedAndCounted` — mutated ts → skipped, `r1_violations=1`, 0 published.

Both pass. Build clean on touched packages.

## Companion

The lake recon (infra) gets an R1 detector — any `event_id` under >1 `timestamp` → MISMATCH alert — to catch anything that bypasses this guard (e.g. a direct producer).

## Summary by CodeRabbit

- **Bug Fixes**
  - Raw event reprocessing now detects timestamp mismatches for existing events.
  - Mismatched events are skipped instead of being published.
  - Processing results and completion logs now report the number of detected violations.

- **Documentation**
  - Clarified event ID uniqueness, timestamp immutability, deduplication, reprocessing behavior, and generated IDs.

- **Tests**
  - Added coverage confirming valid events are published and mismatched events are skipped and counted.

### DIFF
--- a/internal/api/dto/events.go
+++ b/internal/api/dto/events.go
@@ -25,9 +25,18 @@ import (
 
 type IngestEventRequest struct {
 	EventName          string                 `json:"event_name" validate:"required" binding:"required" example:"api_request" csv:"event_name"`
+	// EventID is the caller's idempotency key. CONTRACT: unique per logical event, and IMMUTABLE
+	// once sent — never reuse an event_id for a different event, and never resend the same event_id
+	// with a different timestamp. Events dedup on (event_id, timestamp) in both ClickHouse
+	// (ReplacingMergeTree, ORDER BY ...timestamp,id) and the analytics lake; a reused event_id with
+	// a changed timestamp is treated as a NEW event and double-counts. Reprocessing enforces this
+	// (rejects a timestamp change); recon flags any event_id seen under two timestamps. Omit to have
+	// one generated.
 	EventID            string                 `json:"event_id" example:"event123" csv:"event_id"`
 	CustomerID         string                 `json:"customer_id" example:"customer456" csv:"customer_id"`
 	ExternalCustomerID string                 `json:"external_customer_id" validate:"required" binding:"required" example:"customer456" csv:"external_customer_id"`
+	// Timestamp is the event's business time. IMMUTABLE per event_id (see EventID) — part of the
+	// dedup identity, so changing it for an existing event_id creates a duplicate.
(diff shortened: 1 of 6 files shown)


agrim123:
Closing — not needed for our reprocess model.

Our only reprocess path is Kafka consumer-group offset-rewind: replays the identical raw messages through the same transform, so event_timestamp is reproduced byte-for-byte. R1 (a changed event_timestamp for an existing event_id) cannot arise there — the guard would never fire.

This guard targets the DB-replay path (raw_events_reprocessing.go re-reads raw payloads and re-parses createdAt through possibly-updated transform code), which we won't operate. No runtime double-billing risk on our path, so the guard is dead weight.

Detect-side backstop stays: the recon detector in infrastructure #314 (SELECT id ... HAVING count(DISTINCT timestamp) > 1) still flags any id under two timestamps from any source.

PR #2666, opened 2026-08-26
feat(analytics): publish meter_usage to analytics topic

agrim123:
## What

Rebuilds the meter_usage → analytics Kafka publisher that landed in #2658 and was reverted in #2662, with a new config surface and late-event routing.

After `BulkInsertMeterUsage` commits to ClickHouse, meter_usage records are additively published to an analytics Kafka topic so the analytics plane (CH(B) / lake) can ingest them. **ClickHouse stays authoritative**: the publish is fire-and-forget — errors are logged and swallowed, and the whole path is a no-op unless the feed is enabled — so it can never fail the insert or affect billing.

## Config (new `analytics` section)

| Env | Default | Purpose |
|---|---|---|
| `FLEXPRICE_ANALYTICS_ENABLED` | `false` | Master gate for the **entire** analytics feed. Off ⇒ nil publisher, every path a no-op. |
| `FLEXPRICE_ANALYTICS_METER_USAGE_TOPIC` | `analytics.meter_usage` | Main (on-time) topic. |
| `FLEXPRICE_ANALYTICS_METER_USAGE_LAZY_TOPIC` | `analytics.meter_usage.lazy` | Late-event topic. |
| `FLEXPRICE_ANALYTICS_LATE_THRESHOLD` | `24h` | `ingested_at − timestamp > threshold` ⇒ route to the lazy topic. |

**Routing:** on-time records → main topic; late records (ingested well after their event timestamp) → lazy topic, so downstream can re-rate an already-closed day separately. If the lazy topic is unset, late records fall back to the main topic — never dropped.

`ingested_at` is stamped at publish time when unset, giving the analytics store a version column for its `(id, ingested_at)` dedup.

## Changes

- `internal/config/config.go` — new `AnalyticsConfig` + `Analytics` field on `Configuration`.
- `internal/config/config.yaml` — `analytics:` block with commented defaults.
- `internal/kafka/meter_usage_publisher.go` — rebuilt publisher; gate + late→lazy routing + fire-and-forget.
- `internal/kafka/meter_usage_publisher_test.go` — prior tests + new: on-time→main, late→lazy, empty-lazy→main fallback, threshold boundary.
- `internal/repository/clickhouse/meter_usage.go` — optional variadic publisher; called after the CH commit (nil-safe no-op).
- `internal/repository/factory.go` — optional `MeterUsageAnalyticsPublisher` on `RepositoryParams`.
- `cmd/server/main.go` — `kafka.NewMeterUsagePublisher` fx provider.

[...]

PR #2669, opened 2026-08-27
feat(analytics): make ingested_at producer-authoritative for meter_usage

agrim123:
## What

Make `ingested_at` uniform so CH(A), CH(B), and the analytics/Kafka feed agree on the **producer-authoritative** value.

Today `BulkInsertMeterUsage` omits `ingested_at` from its INSERT column list and relies on the ClickHouse `DEFAULT now64(3)` to fill it at CH(A) insert time. That default is a different clock reading than the one the analytics publisher stamps at publish time, so the sinks disagree on the value.

## Changes (3 files)

- `internal/ee/service/meter_usage_tracking.go` — new `stampIngestedAt(records)` helper stamps `IngestedAt = time.Now().UTC()` on every record that is unset, **once, in UTC**, called just before **both** `BulkInsertMeterUsage` sites. Idempotent — never restamps a record that already carries a value. Stamping lives in the service layer, keeping the repo strictly-DB.
- `internal/repository/clickhouse/meter_usage.go` — `BulkInsertMeterUsage` adds `ingested_at` to the INSERT column list and to the per-row `stmt.Append(...)` binding, writing `record.IngestedAt` **verbatim** instead of letting CH re-default it. No stamping in the repo.
- `internal/domain/events/meter_usage.go` — struct doc updated (was "ingested_at is handled by ClickHouse DEFAULT").

## Why

Prerequisite for the analytics producer PR #2666, and sequenced to land **before** it. `ingested_at` is TRUSTED and PRODUCER-STAMPED: stamped exactly once, in UTC, by the service layer and written verbatim to CH — no CH-side default — so the value is identical across CH(A), CH(B), and Kafka. This PR contains **only** the three prerequisite files; the producer/publisher wiring lands separately in #2666.

## Design axiom

`ingested_at` is stamped exactly once per record, in UTC, and never restamped downstream.

## Test

`go build` + `go test` pass for all touched packages (`internal/repository/clickhouse`, `internal/domain/events`, `internal/ee/service`). No interface signatures changed, so no mock regeneration.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized meter usage ingestion timestamps across ClickHouse and analytics feeds.
  * Applied one shared UTC timestamp to each batch of meter usage records before storage.
  * Preserved existing timestamps while assigning timestamps to records that were missing them.
  * Improved timestamp consistency for both bulk and individual meter usage processing.


internal/ee/service/meter_usage_tracking.go
```diff
@@ -216,6 +216,20 @@ func (s *meterUsageTrackingService) RegisterBulkHandler(router *pubsubRouter.Rou
...
+// carry one, ONCE, before the ClickHouse write. ingested_at is producer-authoritative:
+// stamping it here (service layer) and writing it verbatim — instead of relying on CH's
+// DEFAULT now64(3) — makes the value identical across CH(A), CH(B), and the analytics/Kafka
+// feed. Idempotent: a record that already has an ingested_at is never restamped.
+func stampIngestedAt(records []*events.MeterUsage) {
+	now := time.Now().UTC()
+	for _, record := range records {
+		if record.IngestedAt.IsZero() {
+			record.IngestedAt = now
+		}
```
DiabolusGX:
```suggestion
		record.IngestedAt = now
```

agrim123: Just a safe guard to prevent null timestamp anywhere in downstream

DiabolusGX (approved):

DiabolusGX (approved):

PR #2043, opened 2026-06-20
feat(timezone): add timezone support to invoicing

omkar273:
## Goal

Bill customers in their local timezone, not UTC. A customer in `Asia/Kolkata` (IST, UTC+5:30) on a monthly calendar plan should have period boundaries at **IST midnight** — not UTC midnight (which would be 05:30 IST, invisible and confusing to the customer).

---

### 1. Customer entity — new `timezone` field

- `timezone` (IANA name, e.g. `"Asia/Kolkata"`) added to the `customers` table, defaulting to `"UTC"`
- Accepted on `POST /customers` and `PATCH /customers/:id`
- Validated via `time.LoadLocation()` — invalid names return HTTP 400

### 2. Subscription — timezone inherited from customer

- At subscription creation, `customer_timezone` is always set from the customer record (API caller's value is ignored)
- Stored on the subscription and used for all billing period calculations throughout its lifetime

### 3. `NextBillingDate` — timezone-aware period boundaries

- Renamed from `NextBillingDateWithTimezone`; old positional-arg version removed
- Re-locates `CurrentPeriodStart` and `BillingAnchor` into the customer's IANA timezone before computing the next boundary
- Result is returned as UTC — stored dates are always UTC instants, just computed in local time
- All callers (`subscription.go`, `billing.go`, `creditgrant.go`, `coupon_association.go`, `subscription_trial.go`, `feature_usage_tracking.go`) updated

### 4. `CalculateCalendarBillingAnchor` — calendar boundary in local time

- Previously always computed the "start of next month" as UTC midnight
- Now accepts `timezone string` and computes the boundary at the customer's local midnight
- Example: IST customer starting Feb 15 → calendar anchor is `2024-02-29T18:30:00Z` (IST midnight March 1), not `2024-03-01T00:00:00Z`
- Covers all periods: DAILY, WEEKLY, MONTHLY, QUARTER, HALF_YEAR, ANNUAL

### 5. Period helper functions — all thread timezone

`CalculateBillingPeriods`, `FindPeriodForDate`, `GetNextUsageResetAt`, `CalculatePeriodID` — each gained a `timezone string` parameter and replaced internal `nextBillingDateCore` calls with the timezone-aware `NextBillingDate`

### 6. ClickHouse window functions — timezone-aware bucketing

- `formatWindowSize` and `formatWindowSizeWithBillingAnchor` in `aggregators.go` now accept `tz string`
- UTC: `toStartOfDay(timestamp)` — unchanged for UTC customers
[...]

PR #2671, opened 2026-08-27
fix(creditgrant): compute DURATION expiry in the subscription's timezone

ojasagarwal-flex:
## Problem

Credit grant **period** boundaries have been timezone-aware since #2043, but the `DURATION` branch of the expiry calculation was left doing raw UTC-instant math on `cga.PeriodStart`. The two now disagree for any non-UTC customer.

Monthly grant anchored at `2026-03-01 00:00 IST` (stored as `2026-02-28T18:30Z`):

| | Result |
|---|---|
| Period end (`NextBillingDate`, tz-aware) | `2026-04-01 00:00 IST` ✅ |
| One-month expiry (before this PR) | `2026-03-29 00:00 IST` ❌ — 3 days early |

The customer's credits expire three days before the period they belong to rolls over. `DAY` and `WEEK` added a fixed `24h*n`, which additionally drifts an hour across a DST transition (`America/New_York`, 7-day expiry over spring-forward: `01:00` local instead of `00:00`).

`BILLING_CYCLE` expiry was already threading `subscription.Timezone` — so the bug sat 30 lines above correct code in the same function.

## Change

New `types.AddExpiryDuration(start, duration, unit, timezone)`: relocates the start into the customer's IANA location, advances it with calendar arithmetic, returns a UTC instant — the same convention `NextBillingDate` follows. `AddDate` is used for **every** unit including `DAY`/`WEEK`, so the local wall clock survives DST.

- `applyCreditGrantToWallet`'s `DURATION` branch calls it with `subscription.Timezone`. The 4-case switch collapses to one call; the unknown-unit `ierr` is preserved via the `ok` return.
- `ResolveCreditsExpiry` keeps its signature and delegates with `DefaultTimezone`. Its callers (wallet initial/bonus credits) top up from a request that carries no customer timezone, so **behaviour there is unchanged** — this only de-duplicates the arithmetic.
- `scripts/internal/migrate_cga.go` was passing `""`, so a backfill computed period ends in UTC that the service would then disagree with at runtime. It now resolves the grant's subscription timezone (`subscriptionRepo` was already in scope).

## Scope check

Swept every credit-grant flow for the same class of bug. All three materialization entry points (subscription create, addon attach, plan swap) funnel through `handleCreditGrantsWithStart → CreateCreditGrant → InitializeCreditGrantWorkflow`, which was already tz-aware; rollover, catch-up, cancel and the wallet expiry cron are instant comparisons and therefore timezone-neutral. Also walked every construction of `NextBillingDateParams` / `CalculateBillingPeriodsParams` / `FindPeriodForDateParams` / `GetNextUsageResetAtParams` / `CalculatePeriodIDParams` across `internal/`, `scripts/` and `cmd/` for a missing `Timezone` — none in credit grant code.

**Known gap, deliberately not in this PR:** `types.PreviousBillingDate` has no `Timezone` field at all and shows the same 3-day error. Its only callers are price and entitlement proration on the calendar-billing branch — a separate subsystem that wants its own tests. Happy to raise it separately.

## Tests

`internal/types/creditgrant_expiry_timezone_test.go` — IST month lands on the local 1st, `NY` 7-day and 1-week hold local midnight across spring-forward, unknown unit returns `false`, empty/unresolvable timezone falls back to UTC.

```
go build ./internal/... ./scripts/...            OK
go vet ./internal/ee/service/ ./internal/types/  clean
go test ./internal/...                           green
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Credit grant expiry dates now use the subscription’s local timezone.
  * Day and week durations preserve local time across daylight-saving transitions.
  * Month-based expirations align with local calendar boundaries.
  * Credit grant migrations use subscription timezone data for period-end calculations.
  * Invalid or unavailable timezones safely fall back to UTC.

* **New Features**
  * Entitlement grants now support optional metadata.


migrations/versioned/.hashes
DiabolusGX, line 1: is this intended?

DiabolusGX (dismissed):

DiabolusGX (approved):

PR #2677, opened 2026-08-27
feat(custom-currency): add support for org level custom currency

subratsahilgupta:
## Summary by CodeRabbit

* **New Features**
  * Added support for organization-configured custom currencies.
  * Prices, subscriptions, and wallets now accept configured custom currencies or the organization’s default fiat currency.
  * Invoices can retain custom-currency details while recording fiat totals and conversion rates.
  * Invoice line items now use the applicable custom currency and precision.

* **Bug Fixes**
  * Unconfigured currencies are rejected during price, subscription, and wallet creation.
  * Invoice finalization now consistently converts totals to fiat and preserves the applicable conversion rate.


internal/ee/service/invoice.go
```diff
@@ -279,6 +280,21 @@ func (s *invoiceService) CreateEmptyDraftInvoice(ctx context.Context, req dto.Cr
...
+			return err
+		}
+		inv.Currency = currency
+
+		fiatCurrency, err := ccSvc.ResolveFiatCurrency(txCtx, inv.Currency, req.TargetCurrency)
+		if err != nil {
+			return err
+		}
+		if fiatCurrency != "" {
+			inv.TargetCurrency = &types.TargetCurrency{FiatCurrencyCode: fiatCurrency}
```
DiabolusGX: I don't think we immediately need to store `targetCurrency` when generating invoice.
It's only required when interacting with external systems like sync invoice to zoho or take payment for invoice.

Shouldn't resolve to target fiat currency at that time?

internal/types/settings.go
```diff
@@ -34,6 +34,7 @@ const (
 	SettingKeyDraftInvoiceRecomputeConfig SettingKey = "draft_invoice_recompute_config"
 	SettingKeySAMLConfig                  SettingKey = "saml_config"
 	SettingKeyWalletTopupConfig           SettingKey = "wallet_topup_config"
+	SettingKeyOrgCustomCurrencyConfig     SettingKey = "org_custom_currency_config"
```
DiabolusGX: Please rename at other places as well. Don't think we should call it "org currency"

```suggestion
	SettingKeyCustomCurrencyConfig     SettingKey = "custom_currency_config"
```

internal/ee/service/invoice.go
```diff
@@ -977,6 +1008,24 @@ func (s *invoiceService) performFinalizeInvoiceActions(ctx context.Context, inv
 			return ierr.NewError("invoice is not in draft status").WithHint("invoice was finalized concurrently").Mark(ierr.ErrValidation)
 		}
 
+		// Freeze the rate first so every fiat column below uses the sealed rate.
+		if cc := lockedInv.CustomCurrency; cc != nil {
+			settingsSvc := NewSettingsService(s.ServiceParams).(*settingsService)
+			ccCfg, err := GetSetting[types.CustomCurrencyConfig](settingsSvc, txCtx, types.SettingKeyCustomCurrencyConfig)
+			if err != nil {
+				return err
```
Harshit-1104, line 1026: caching to validate

internal/ee/service/invoice.go
```diff
@@ -995,21 +1044,32 @@ func (s *invoiceService) performFinalizeInvoiceActions(ctx context.Context, inv
...
+				// Recalculate total with credits applied, in the ledger currency
+				if cc := lockedInv.CustomCurrency; cc != nil {
+					cc.Total = cc.Subtotal.Sub(cc.TotalDiscount).Sub(cc.TotalPrepaidCreditsApplied)
+					if cc.Total.IsNegative() {
+						cc.Total = decimal.Zero
+					}
+					cc.AmountDue = cc.Total
+					lockedInv.ProjectCustomCurrency()
+					if err := s.persistProjectedLineItems(txCtx, lockedInv); err != nil {
+						return err
```
Harshit-1104: validate if needed?

internal/ee/service/invoice.go
```diff
@@ -4366,10 +4488,14 @@ func (s *invoiceService) GetInvoiceWithBreakdown(ctx context.Context, req dto.Ge
...
 func (s *invoiceService) recalculateInvoiceTotals(inv *dto.InvoiceResponse) {
-	// Calculate new subtotal from line item amounts
+	// Sum in the ledger currency and convert once; summing fiat lines would drift.
 	newSubtotal := decimal.Zero
 	for _, lineItem := range inv.LineItems {
-		newSubtotal = newSubtotal.Add(lineItem.Amount)
+		newSubtotal = newSubtotal.Add(lineItem.Ledger().Amount)
+	}
+	if inv.CustomCurrency != nil {
+		inv.CustomCurrency.Subtotal = newSubtotal
```
Harshit-1104, line 4582: use projection everywhere

internal/types/custom_currency.go
```diff
@@ -0,0 +1,165 @@
...
+// custom currency or the default fiat. Unconfigured tenants are unaffected.
+func (c CustomCurrencyConfig) EnforceCurrency(currency string) error {
+	if len(c.CustomCurrencies) == 0 {
+		return nil
+	}
+	currency = strings.ToLower(currency)
+	if c.IsCustom(currency) || currency == c.DefaultFiatCurrency {
+		return nil
+	}
+
```
Harshit-1104, line 175: handle to check if fiat exists in custom currency mappings, if it does, then allow

internal/domain/invoice/model.go
```diff
@@ -291,6 +295,81 @@ func (i *Invoice) Validate() error {
...
+			PrepaidCreditsApplied: item.PrepaidCreditsApplied,
+		}
+	}
+
+	i.ProjectCustomCurrency()
+}
+
+// MirrorTaxIntoLedger restates the totals tax moved. Tax is computed and stored in
+// fiat, so it has no ledger form of its own.
+func (i *Invoice) MirrorTaxIntoLedger() {
```
Harshit-1104: check if ProjectCustomCurrency could be used

internal/domain/invoice/line_item.go
```diff
@@ -107,6 +112,42 @@ func (i *InvoiceLineItem) FromEnt(e *ent.InvoiceLineItem) *InvoiceLineItem {
...
+	return types.CustomCurrencyLineItem{
+		Amount:                i.Amount,
+		LineItemDiscount:      i.LineItemDiscount,
+		InvoiceLevelDiscount:  i.InvoiceLevelDiscount,
+		PrepaidCreditsApplied: i.PrepaidCreditsApplied,
+	}
+}
+
+// SetLedgerPrepaidCreditsApplied writes back into whichever set Ledger reads from.
+func (i *InvoiceLineItem) SetLedgerPrepaidCreditsApplied(applied decimal.Decimal) {
```
Harshit-1104: clean this up, use ProjectCustomCurrency

internal/ee/service/credit_adjustment.go
```diff
@@ -302,8 +310,10 @@ func (s *creditAdjustmentService) ApplyCreditsToInvoice(ctx context.Context, inv
 		// We calculated these values earlier, now we're just saving them to the database
 		totalAmountApplied := decimal.Zero
 		for _, lineItem := range inv.LineItems {
-			if lineItem.PrepaidCreditsApplied.GreaterThan(decimal.Zero) {
-				totalAmountApplied = totalAmountApplied.Add(lineItem.PrepaidCreditsApplied)
+			applied := lineItem.Ledger().PrepaidCreditsApplied
+			lineItem.ProjectCustomCurrency(inv.CustomCurrency, inv.Currency)
```
Harshit-1104, line 314: move just below update

internal/ee/service/credit_adjustment.go
```diff
@@ -315,11 +325,15 @@ func (s *creditAdjustmentService) ApplyCreditsToInvoice(ctx context.Context, inv
 		// persisted to the database. The caller is responsible for updating the invoice
 		// in the database if they need to persist TotalPrepaidApplied.
 		// This design allows callers to batch invoice updates with other operations.
-		inv.TotalPrepaidCreditsApplied = totalAmountApplied
+		if inv.CustomCurrency != nil {
+			inv.CustomCurrency.TotalPrepaidCreditsApplied = totalAmountApplied
+		} else {
+			inv.TotalPrepaidCreditsApplied = totalAmountApplied
+		}
```
Harshit-1104, line 332: project

Harshit-1104 (changes requested):

internal/domain/invoice/model.go
```diff
@@ -291,6 +295,82 @@ func (i *Invoice) Validate() error {
...
+	i.TotalDiscount = cc.ToFiat(cc.TotalDiscount, i.Currency)
+	i.TotalTax = cc.ToFiat(cc.TotalTax, i.Currency)
+	i.TotalPrepaidCreditsApplied = cc.ToFiat(cc.TotalPrepaidCreditsApplied, i.Currency)
+	i.Total = cc.ToFiat(cc.Total, i.Currency)
+	i.AmountDue = cc.ToFiat(cc.AmountDue, i.Currency)
+
+	i.AmountRemaining = i.AmountDue.Sub(i.AmountPaid)
+	if i.AmountRemaining.IsNegative() {
+		i.AmountRemaining = decimal.Zero
+	}
```
Harshit-1104: this logic shouldn't be in ProjectCustomCurrency

internal/ee/service/invoice.go
```diff
@@ -3769,6 +3873,43 @@ func (s *invoiceService) RecalculateInvoice(ctx context.Context, id string) (*dt
 }
 
 // RecalculateTaxesOnInvoice recalculates taxes on an invoice if it's a subscription invoice.
+// persistProjectedLineItems writes line items whose fiat amounts were just projected
```
Harshit-1104: the comment above this is not related to this function

Harshit-1104 (changes requested):

internal/domain/invoice/model.go
subratsahilgupta: Tax is the one amount that is produced in fiat rather than in the custom currency, so it cannot be captured, only converted back.

internal/domain/invoice/line_item.go
subratsahilgupta: Credits are drawn in the denomination currency, so the applied amount has to land in the field denominated the same way.
Capture cannot replace it here either. Credit adjustment runs at invoice.go:1060, after the projection at :1037, so the plain fields already hold fiat. Capturing at that point would copy the projected fiat amounts back into the denomination and then project them a second time.

internal/ee/service/credit_adjustment.go
subratsahilgupta: not needed

internal/domain/invoice/line_item.go
```diff
@@ -107,6 +112,43 @@ func (i *InvoiceLineItem) FromEnt(e *ent.InvoiceLineItem) *InvoiceLineItem {
...
+
+// SetDenominationPrepaidCreditsApplied writes the applied credits into the field whose
+// currency they were drawn in. Its fiat counterpart follows from ProjectCustomCurrency.
+func (i *InvoiceLineItem) SetDenominationPrepaidCreditsApplied(applied decimal.Decimal) {
+	if i.CustomCurrency != nil {
+		i.CustomCurrency.PrepaidCreditsApplied = applied
+		return
+	}
+	i.PrepaidCreditsApplied = applied
+}
```
Harshit-1104: Cleanup of this to be picked in enhancement as discussed

Harshit-1104 (approved):

internal/ee/service/credit_adjustment.go
```diff
@@ -298,16 +305,16 @@ func (s *creditAdjustmentService) ApplyCreditsToInvoice(ctx context.Context, inv
...
-		// We calculated these values earlier, now we're just saving them to the database
+		// Total up what was applied. The line items themselves are persisted by the
+		// caller, once the amounts have been converted.
 		totalAmountApplied := decimal.Zero
 		for _, lineItem := range inv.LineItems {
-			if lineItem.PrepaidCreditsApplied.GreaterThan(decimal.Zero) {
-				totalAmountApplied = totalAmountApplied.Add(lineItem.PrepaidCreditsApplied)
-				if err := s.InvoiceLineItemRepo.Update(ctx, lineItem); err != nil {
-					return err
-				}
```
Harshit-1104, line 309: i think this was from before, old code, validate once, shouldn't be removed

Harshit-1104 (changes requested):

internal/ee/service/credit_adjustment.go
subratsahilgupta: I checked it 
so during finalization we are updating repo
purposefully moved

With RestoreFromDenomination running first, lineItem.PrepaidCreditsApplied holds credits during credit adjustment. The old Update there would have written custom currency value into the fiat column , the value belongs in custom_currency.prepaid_credits_applied, and the fiat column should hold fiat value right.

So the write has to happen after capture+project, which is the caller's job. Credit adjustment now only computes and totals.

Harshit-1104 (approved):

DiabolusGX (approved):

PR #2688, opened 2026-08-29
feat(migrations): run versioned SQL migrations on deploy

ojasagarwal-flex:
Replaces Ent auto-migration as the PostgreSQL deploy mechanism. Schema changes ship as reviewed `.sql` files; dbmate applies what a database has not recorded and writes it to `schema_migrations`.

## Why

Ent auto-migration inferred DDL by diffing the live schema against the Go models at deploy time. Nothing reviewed what it would run and nothing recorded what it did, so the same code produced different DDL against different databases. It also skipped `DropIndex`, `DropColumn` and `ModifyIndex` — which is how a wrong unique index survived on India production from 2026-07-25 unnoticed.

## Single migration folder

The Ent-generated baseline is now the first file in `migrations/versioned/postgres/`. A fresh install builds the whole schema from that folder and needs nothing else.

An **existing** database is adopted once — its versions are recorded, zero DDL executes — after which only new migrations run there. That is deliberate: the existing files describe history each deployment already lived through by its own route (AutoMigrate, DMS into AlloyDB, hand-run DDL), and replaying it would be wrong.

Forgetting to adopt is safe. The baseline collides on its first `CREATE TABLE` and stops with an empty ledger, rather than applying later migrations to a schema nobody checked.

`adopt.sh` gains `--dry-run`, since adoption is run by hand once per deployment and per client, against production.

## `migrate postgres up`

New subcommand on the existing binary. It refuses two states rather than proceeding into them:

- **tables but no ledger** — never adopted; prints the `make migrate-adopt` commands
- **any INVALID index** — `CREATE INDEX CONCURRENTLY IF NOT EXISTS` skips an invalid index silently, so continuing would record the migration as applied and leave the index broken forever. Prints the `DROP INDEX CONCURRENTLY` for each.

It sets `lock_timeout=3s` and `statement_timeout=0` on the connection. Neither can live in the SQL: a `transaction:false` file may contain only one statement, leaving no room for a `SET`.

dbmate ships as a **binary** built in an isolated Dockerfile stage, not a Go dependency — its module pulls in gorm, BigQuery, MySQL and SQLite drivers the application has no use for. `go.mod` is untouched.

## Deployment

**Helm** — `migration.steps.dbmate` replaces `migration.steps.ent` in the existing pre-upgrade hook Job. Enabling both fails rendering rather than running the schema through two mechanisms. The Ent path stays in the image, so rollback is a values flip.

`backoffLimit` now defaults to `0`: retrying a killed migration is what turns an aborted index build into a permanently invalid one. The template no longer routes it through `| default 3`, which treats `0` as empty and would have silently restored 3.

**ECS** — a `migrate` job runs one task per target before the canary and before the production approval gate. The task is built from the API service's own task definition, so it inherits the same secrets, role, subnets and security groups the app uses. Migrations land before the new code, which is safe because migrations are additive by rule.

## CI

- dbmate pinned to the version the image builds, instead of `releases/latest`
- new gate runs the deploy command itself: fresh install must succeed; an unadopted database must be refused without creating a ledger

## Verified

| | |
|---|---|
| Fresh DB | 7 applied, 56 tables, ledger 7 |
| Re-run | no-op, exit 0 |
| Existing unadopted | refused, ledger empty, exit 1 |
| INVALID index present | refused, names the index and the DROP |
| Inside the built image | `dbmate` on PATH, migrations present, all of the above |
| Helm: default / rollback / both-true / disabled | dbmate step / ent step / render fails / no Job |
| `lock_timeout` + `statement_timeout` | confirmed applied via `SHOW` |

## Before merging this reaches a database

Every existing deployment must be adopted before it takes chart 1.4.0. **GCP staging is done; India prod, US prod and client databases are not.**

Any values file setting `migration.steps.ent: true` explicitly must drop it, or rendering fails on the mutual-exclusion guard.

## Known, unchanged

- ClickHouse versioned set stays inert; its gate remains disabled
- Sarvam still runs migrations by hand via `run.sh`; switching it to `up` is an infra-repo change
- No drift monitoring — nothing reports a region sitting N migrations behind

## Summary by CodeRabbit

- **New Features**
  - Added reviewed, versioned PostgreSQL migrations for fresh databases and upgrades.
  - Added migration adoption with dry-run previews and optional version targeting.
  - Added automated PostgreSQL and Kafka migrations before deployments.
  - Added optional Kafka ACL seeding and packaged migration tooling.

- **Bug Fixes**
  - Deployments now stop when migrations fail.
  - Preserved explicit retry settings, including zero retries.
  - Added safeguards against conflicting migration methods and unsafe database states.

- **Documentation**
  - Updated upgrade, adoption, migration deadline, and rollback guidance.


.github/actions/run-ecs-migration/action.yml
```diff
@@ -0,0 +1,234 @@
...
+# A task registered from the API service's own task definition inherits its
+# execution role, secrets, subnets and security groups, so the migration reaches
+# the database by exactly the same path the application does — no second copy of
+# the connection configuration to drift.
+#
+# The task runs `./migrate postgres up`, which applies the reviewed .sql files in
+# migrations/versioned/postgres and records them in schema_migrations. It refuses
+# to touch a database that has tables but no ledger; adopt it first (see
+# migrations/versioned/README.md).
+name: 'Run ECS migration task'
```
agrim123, line 15: Instead of triggering a ecs task from here, can't we make this part of CodePipeline? safer

agrim123 (changes requested):

agrim123 (approved):

agrim123 (dismissed):

agrim123 (approved):

PR #2748, opened 2026-09-03
fix(customer-portal): let an auto top-up cooloff be cleared

dskhairnar:
A cooloff could be **set but never reset**. Turning it off saved without error and left the old value in place.

## Cause

`UpdateWallet` merges auto top-up field by field and only reads the cooloff when the pointer is non-nil:

```go
if req.AutoTopup.Cooldown != nil {
    if req.AutoTopup.Cooldown.IsEmpty() { current.Cooldown = nil } else { current.Cooldown = req.AutoTopup.Cooldown }
}
```

Go unmarshals JSON `null` into a **nil pointer**, which is indistinguishable from the field being absent. The portal passed the customer's cleared cooloff straight through, so the block was skipped entirely and the stored value survived. The save returned 200 — which is why it looked like it had worked.

## Fix

The portal submits the whole auto top-up form, so a cooloff the customer did not set means there is none. `UpdateAutoTopup` now says that explicitly, translating an absent or empty cooloff into the merge's clear signal rather than relying on a nil pointer to carry the intent.

`AutoTopup.Validate` also no longer rejects an empty cooloff. That validation is not wired into the update path today — but it is the obvious hardening for someone to add later, and it would have made clearing impossible again, since `Duration.Validate` refuses a value of zero. Removing that landmine now costs one condition.

## Tests

- `TestAutoTopupCooloffCanBeCleared` — stores a cooloff, then clears it with an absent one. **Verified it fails without the service change** and passes with it.
- `TestAutoTopupCooloffClearsOnZeroDuration` — the same clear said the other way, with an explicitly zero-valued duration. This path already worked; the test documents it so it stays working.

## Relationship to the frontend

[flexprice-front#1354](https://github.com/flexprice/flexprice-front/pull/1354) works around this from the client by sending `{value: 0}` instead of `null`. With this merged, that workaround is no longer needed and the portal can send whichever it prefers — both now clear. **I'd suggest closing #1354 in favour of this**, since fixing it here also covers every other client of the endpoint.

## Verification

- `go build ./internal/...`, `go vet` — clean
- `go test ./internal/ee/service/ -run TestPortalWalletSuite` — passing
- `go test ./internal/types/...` — passing
- `gofmt -l` — clean
- `make lint-ci` — clean (exit 0)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed auto top-up settings so clearing the cooldown correctly removes a previously saved cooldown.
  * Empty or zero-duration cooldown values now clear the saved setting.
  * Partial auto top-up updates preserve existing settings that are not included in the request.
  * Improved reliability when updating auto top-up settings with only selected fields.

### DIFF
--- a/internal/ee/service/customer_portal_wallet.go
+++ b/internal/ee/service/customer_portal_wallet.go
@@ -126,26 +126,7 @@ func (s *customerPortalService) UpdateAutoTopup(ctx context.Context, walletID st
 
 	// Invoicing is pinned, never taken from the request: it selects the transaction
 	// reason and so decides whether credits are paid for at all.
-	// The portal submits the whole auto top-up form, so a cooloff the customer did
-	// not set means there is none — and that has to be said explicitly.
-	//
-	// UpdateWallet merges auto top-up field by field and only reads the cooloff
-	// when the pointer is non-nil. A JSON null unmarshals to nil, which is
-	// indistinguishable from the field being absent, so passing it straight through
-	// left the stored cooloff untouched: it could be set but never reset. An empty
-	// Duration is the merge's clear signal.
-	cooldown := req.Cooldown
-	if cooldown.IsEmpty() {
-		cooldown = &types.Duration{}
-	}
-
-	autoTopup := &types.AutoTopup{
-		Enabled:   lo.ToPtr(req.Enabled),
-		Threshold: req.Threshold,
-		Amount:    req.Amount,
-		Invoicing: lo.ToPtr(true),
-		Cooldown:  cooldown,
-	}
+	autoTopup := types.NewAutoTopup(req.Enabled, req.Threshold, req.Amount, true, req.Cooldown)
 
 	if req.Enabled {
 		if req.Amount == nil || req.Threshold == nil {
--- a/internal/ee/service/customer_portal_wallet.go
+++ b/internal/ee/service/customer_portal_wallet.go
@@ -126,12 +126,25 @@ func (s *customerPortalService) UpdateAutoTopup(ctx context.Context, walletID st
 
 	// Invoicing is pinned, never taken from the request: it selects the transaction
 	// reason and so decides whether credits are paid for at all.
+	// The portal submits the whole auto top-up form, so a cooloff the customer did
+	// not set means there is none — and that has to be said explicitly.
+	//
+	// UpdateWallet merges auto top-up field by field and only reads the cooloff
+	// when the pointer is non-nil. A JSON null unmarshals to nil, which is
+	// indistinguishable from the field being absent, so passing it straight through
+	// left the stored cooloff untouched: it could be set but never reset. An empty
+	// Duration is the merge's clear signal.
+	cooldown := req.Cooldown
+	if cooldown.IsEmpty() {
+		cooldown = &types.Duration{}
+	}
+
 	autoTopup := &types.AutoTopup{
 		Enabled:   lo.ToPtr(req.Enabled),
 		Threshold: req.Threshold,
 		Amount:    req.Amount,
 		Invoicing: lo.ToPtr(true),
-		Cooldown:  req.Cooldown,
+		Cooldown:  cooldown,
 	}
 
 	if req.Enabled {
--- a/internal/ee/service/customer_portal_wallet_test.go
+++ b/internal/ee/service/customer_portal_wallet_test.go
@@ -228,6 +228,50 @@ func (s *PortalWalletSuite) TestAutoTopupEnableRequiresChargeableMethod() {
 	s.False(ierr.IsValidation(err), "a missing card is a state conflict, not a bad request")
 }
 
+// A cooloff has to be resettable, not just settable. UpdateWallet only reads the
+// cooloff when the pointer is non-nil, and a JSON null unmarshals to nil — which is
+// indistinguishable from an absent field — so passing it through left the stored
+// cooloff in place and the customer could never clear it.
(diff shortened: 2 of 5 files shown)


internal/ee/service/customer_portal_wallet.go
```diff
@@ -126,12 +126,25 @@ func (s *customerPortalService) UpdateAutoTopup(ctx context.Context, walletID st
...
+		cooldown = &types.Duration{}
+	}
+
 	autoTopup := &types.AutoTopup{
 		Enabled:   lo.ToPtr(req.Enabled),
 		Threshold: req.Threshold,
 		Amount:    req.Amount,
 		Invoicing: lo.ToPtr(true),
-		Cooldown:  req.Cooldown,
+		Cooldown:  cooldown,
```
Harshit-1104: remove ai comments and create a NewAutoTopup constructor and reuse in this and other place
this logic is to be in one place

internal/types/wallet.go
```diff
@@ -384,8 +384,14 @@ func (a *AutoTopup) Validate() error {
 			WithHint("Invoicing boolean is required").
 			Mark(ierr.ErrValidation)
 	}
-	if err := a.Cooldown.Validate(); err != nil {
-		return err
+	// An empty cooloff is how a caller asks for the stored one to be cleared, so it
+	// must not be rejected as a malformed duration. Without this, wiring this
+	// validation into the update path — a reasonable hardening — would make
+	// clearing a cooloff impossible again.
+	if !a.Cooldown.IsEmpty() {
```
Harshit-1104: not needed

internal/ee/service/customer_portal_wallet.go
dskhairnar: Done — the cooloff normalisation now lives in a `NewAutoTopup` constructor in `internal/types/wallet.go`, and the portal service just calls it. The inline explanation came out with it; only the "invoicing is pinned" note remains, since that one is not obvious from the code.

Worth flagging: `UpdateAutoTopup` is the only place that builds an `AutoTopup` from a request. The other `&types.AutoTopup{}` in `wallet.go` is an empty accumulator the merge initialises when the wallet has no config yet, so routing it through the constructor would set fields the merge is about to overwrite. Happy to pull that in too if you meant that one.

internal/types/wallet.go
dskhairnar: Removed. You're right that it isn't needed — `AutoTopup.Validate` is only reached from `CreateWalletRequest.Validate`, never from the update path this fix concerns, so the guard was defensive rather than load-bearing. `Validate` is back to validating the cooloff unconditionally.

internal/ee/service/customer_portal_wallet_test.go
```diff
@@ -228,6 +228,50 @@ func (s *PortalWalletSuite) TestAutoTopupEnableRequiresChargeableMethod() {
...
+}
+
+// The same clear, said the other way: an explicitly zero-valued duration.
+func (s *PortalWalletSuite) TestAutoTopupCooloffClearsOnZeroDuration() {
+	s.connect(types.SecretProviderChargebee)
+
+	_, err := s.svc.UpdateAutoTopup(s.ctx, s.walletID, &dto.PortalUpdateAutoTopupRequest{
+		Enabled:  false,
+		Cooldown: &types.Duration{Value: 30, Unit: types.DurationUnitMinute},
+	})
```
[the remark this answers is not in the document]
dskhairnar: Fixed — the setup call's response is now captured and asserted (cooloff present, value 30) before the clear runs, so a persistence regression can't let the clear pass against an already-empty value.

internal/ee/service/customer_portal_wallet.go
dskhairnar: Follow-up — you meant the `UpdateWallet` path, and you were right that it was the duplication that mattered. Done in `7587e6d`.

`AutoTopup` now has a builder in `internal/types`, following the `entitlementgrant` convention, and the ~20-line merge block in `UpdateWallet` collapses to:

```go
existing.AutoTopup = types.NewAutoTopupBuilder(existing.AutoTopup).
    WithAutoTopup(req.AutoTopup).
    Build()
```

So there is one place that knows the rules now: setters ignore a nil field, and `WithCooldown` treats an empty `Duration` as a clear. The portal keeps using `NewAutoTopup` for its full-form submission, which normalises an absent cooloff into that clear signal.

One behaviour change worth flagging: the old block did `current := existing.AutoTopup` and mutated it in place, so a rejected update left the loaded entity half-changed. The builder copies first.

Covered by `internal/types/wallet_test.go` — I checked the tests fail if the empty-Duration rule is removed, rather than just passing.

internal/types/wallet.go
```diff
@@ -368,6 +368,104 @@ type AutoTopup struct {
...
+	return &AutoTopup{
+		Enabled:   lo.ToPtr(enabled),
+		Threshold: threshold,
+		Amount:    amount,
+		Invoicing: lo.ToPtr(invoicing),
+		Cooldown:  cooldown,
+	}
+}
+
+// autoTopupBuilder copies an existing config and applies field updates.
```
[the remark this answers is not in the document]
dskhairnar: Agreed — removed in `e28c5be`, along with two others that restated their signatures. The comments that stay are the ones covering the nil-vs-empty `Duration` rule, which is the actual edge case here.

Harshit-1104 (dismissed):

Harshit-1104 (approved):

PR #2520, opened 2026-08-11
Polling api

Harshit-1104:
## **User description**
Fixes #<issue-number>

## 📝 Description
Please provide a clear and concise description of what this PR does.

## 🔨 Changes Made
- [ ] Feature 1
- [ ] Bug Fix
- [ ] Refactor
- [ ] Documentation Update

## ✅ Checklist
- [ ] My code follows the project's code style.
- [ ] I have added tests where applicable.
- [ ] All new and existing tests pass.
- [ ] I have updated documentation (if needed).

## 📷 Screenshots / Demo (if applicable)
_Add screenshots or demo links here._

## Summary by CodeRabbit

* **New Features**
  * Zoho invoice synchronization now supports item-level discounts, including pre-tax discount handling.
  * Discounts are accurately distributed across invoice lines with currency rounding and capped values.
* **Documentation**
  * Added a design specification for checkout-session polling and reconciliation.
  * Documented checkout status responses, payment details, polling guidance, stale-state indicators, expiration handling, and available payment actions.
  * Clarified concurrency safeguards, webhook interactions, cleanup behavior, and graceful handling of gateway failures.

___

## **CodeAnt-AI Description**
Document checkout polling and payment reconciliation behavior

### What Changed
- Documents polling the existing checkout-session endpoint to reconcile pending payments with the payment provider when webhooks are delayed or missing
- Defines response fields for terminal status, payment details, polling backoff, stale results, and payment actions
- Describes retry handling, expiry grace periods, final payment checks, refunds, draft cleanup, and safeguards against duplicate completion
- Records proposed cancellation behavior and known payment lifecycle issues that affect checkout recovery

### Impact
`✅ Fewer checkout sessions stuck after successful payment`
`✅ Clearer polling and stale-state feedback for clients`
`✅ Safer recovery for late payments and expired checkouts`
<details><summary><strong>💡 Usage Guide</strong></summary>

### Checking Your Pull Request
Every time you make a pull request, our system automatically looks through it. We check for security issues, mistakes in how you're setting up your infrastructure, and common code problems. We do this to make sure your changes are solid and won't cause any trouble later.

### Talking to CodeAnt AI
Got a question or need a hand with something in your pull request? You can easily get in touch with CodeAnt AI right here. Just type the following in a comment on your pull request, and replace "Your question here" with whatever you want to ask:
<pre>
<code>@codeant-ai ask: Your question here</code>
</pre>
[...]

PR #2760, opened 2026-09-03
feat(checkout): reconcile checkout sessions against the provider on read

ojasagarwal-flex:
Implements the polling design from #2520.

## Problem

A customer pays on a provider-hosted page and the only thing that tells us is the provider's webhook. When it is late, dropped, or errors, the session stays `pending`, the draft subscription never activates, and the customer has paid for nothing — until the sweeper expires the session and, if a webhook ever arrives, refunds them.

`GET /v1/checkout/sessions/{id}` returned stored state and never consulted the gateway, so polling it returned the same stale answer forever.

## Approach

The same GET now reconciles a non-terminal session against the provider when a debounce window has elapsed, and routes any transition through the existing `CompleteCheckoutSession`. Same route, same verb, no new endpoint; the client asserts nothing.

```
GET /v1/checkout/sessions/{id}
  ├─ session terminal?          → stored state, no gateway call
  ├─ no checkout_payment_id?    → stored state
  ├─ debounce lock unavailable? → stored state, stale = true
  ├─ ask the provider: did this pay?
  ├─ paid? → CompleteCheckoutSession  (ErrAlreadyExists = success)
  └─ + terminal / stale / next_poll_after_ms / payment
```

## The finding that shaped it

**Checkout payments carried no gateway identifiers.** `CreatePaymentForCheckout` runs before the provider is contacted and nothing backfilled the row, so `gateway_tracking_id` and `gateway_payment_id` were NULL for the life of a pending checkout. `syncPaymentStatusFromGateway` reads exactly those columns, so it was dead code for every checkout payment on every path — `GET /v1/payments/{id}` silently did nothing for them.

`recordGatewayHandles` now writes the provider ids onto the payment, and the poll reads the handle from there.

## Provider-agnostic

`FetchPaymentState` was added to the existing `interfaces.CheckoutProvider`. Each adapter interprets its own ids — only Razorpay knows `plink_`, `inv_` and `order_` are three objects answered by three endpoints. Chargebee returns `ErrNotImplemented`, which the poll treats as "no information" rather than failure; enabling it later is one adapter method.

## Rules the poll follows

- **Gates on session status, never payment status.** The session is the authority on "we are still trying".
- **Success transitions only.** A declined payment is deliberately left `PENDING` — the link is still live and the customer may retry. Writing `FAILED` would close that window and, since `paymentStatusTransitions[FAILED] = {FAILED}`, then block the successful retry from reaching `SUCCEEDED`.
- **Never fails the read.** Gateway errors are swallowed; 200 with `stale = true`.
- **Redis bounds cost, Postgres guarantees correctness.** The debounce limits gateway calls; `MarkCompleted`'s conditional UPDATE is what applies effects at most once.

## Also in here

- **`expire_by` is sent to Razorpay and read back**, so the link closes before the session and the grace window only covers a payment already in flight. `LinkExpiry` leaves headroom over Razorpay's 15-minute floor — it is evaluated on receipt and rejects a timestamp computed to land exactly on it.
- **`MergeOnto` stops completion clobbering `provider_result`.** Callers pass a fragment; writing it straight to the column dropped `next_action` and `provider_session_id`. Confirmed in production data.
- **Responses carry `terminal`, `stale`, `next_poll_after_ms` and a `payment` block** — on create as well as read. Create is where a client gets the URL, so a zero interval there would stop it before it began.
- **The checkout response no longer leaks `status`, `tenant_id`, `created_by`, `updated_by`** from the embedded domain model. ⚠️ Breaking for any consumer reading them.
- **`CheckoutStatus.IsTerminal()` and `ActiveCheckoutStatuses()`** replace the terminal set that was retyped by hand at six call sites.
- **The customer portal read polls too**, and returns the same hints.

## Timings

| | Razorpay | Chargebee |
| --- | --- | --- |
| `LinkExpiry` | 20 min | 30 min |
| `SessionGrace` | 5 min | 5 min |
| `SessionExpiry` | 25 min | 35 min |

Backoff 2s / 5s / 10s by session age, 0 when terminal. The same value is the debounce TTL, so the advertised cadence and the window cannot drift.

## Testing

26 tests in `checkout_session_poll_test.go`: recovery, terminal short-circuit, 8-way concurrent completion, debounce, fail-open, gateway errors, the poll's limits, handle recording, `provider_result` preservation, expiry ordering.

**Verified end to end against Razorpay test mode with no webhook configured.** A paid session that had not moved was completed by a single read: payment `SUCCEEDED` with `pay_` backfilled, invoice `FINALIZED`, subscription **active**, `provider_result` intact. Also confirmed: debounce returns `stale: true`, a completed session makes zero gateway calls, an unavailable provider returns **200** not 5xx, a decline followed by a retry still completes, and the mandate path produces an `inv_` handle that polls correctly.

Two bugs were found only by running it — the `expire_by` boundary and the missing polling hints on create. Neither was reachable by unit tests.

## Notes for review

- `checkoutSessionService.checkoutProviderFor` is a test seam, nil in production. `*integration.Factory` is concrete with 15 methods; an interface for it is its own refactor.
- `TestConcurrentPolls_CompleteExactlyOnce` trips `-race` inside `testutil.InMemorySubscriptionStore.GetWithLineItems`, which writes to the shared stored pointer. A flaw in the test double, not the path under test — needs a decision if CI runs race.
- `PaymentMethodType` is deliberately untouched. It is hardcoded to `PAYMENT_LINK` and wrong for the auto-charge path, but `handlePaymentFailed` currently depends on that, so the field and the branch must move together. Separate ticket.

## Findings for #2520

- **Bugs #1, #2 and #8 are stale** — including all three the design names as its dependencies. #1 is already fixed by `33fd2dda7`; #2 is unreachable on checkout because `handlePaymentFailed` returns early and records an attempt; #8 is inverted (`TrackAttempts: true`).
- **§5.1's grace window did not exist** — no `expire_by` was sent and no provider expiry read back, so the tightening was dead code and the session could close before the link. Implemented here.
- **§8's table lists `payment → FAILED` as a poll capability**, contradicting its own "the poll gains exactly two capabilities".
- **§3's flowchart draws `WH --> SYNC`** — the webhook has the status in its own payload.
- **§7 is missing**; the document jumps §6 → §8.
- **Portal polling is absent from the PRD**, but the portal calls this endpoint.

## Separately, not in this PR

`cleanupCheckoutSession` reads only `session.Result`, but four of the five flows store their ids in `session.Configuration` — so it under-archives draft invoices and payments for those. Confirmed live. Own ticket.

## Summary by CodeRabbit

- **New Features**
  - Checkout sessions now automatically reconcile payment status with supported providers when viewed.
  - Added polling guidance, including terminal, stale, and next-poll timing indicators.
  - Payment details and gateway tracking information are now surfaced for improved reconciliation.
  - Razorpay checkout links honor expiration times and support invoice, order, payment, and link status checks.
  - Customer portal checkout responses now include terminal and polling state.
  - Checkout expiration now includes provider-specific timing and a grace period.

- **Bug Fixes**
  - Improved handling of delayed, missing, or failed payment notifications.
  - Checkout completion now correctly handles provider updates and payment races.
  - Payment gateway tracking identifiers can advance initiated payments to pending status.

### DIFF
--- a/internal/types/checkout_configuration.go
+++ b/internal/types/checkout_configuration.go
@@ -279,6 +279,48 @@ type CheckoutProviderResult struct {
 	ProviderMetadata map[string]string `json:"provider_metadata,omitempty"`
 }
 
+// MergeOnto returns a copy of base with the non-empty fields of r overlaid. Callers
+// that learn only part of the result — a webhook knowing the payment id but not the
+// redirect action — pass a fragment; writing it straight to the column would drop the
+// NextAction and ProviderSessionID recorded at link creation. Either side may be nil.
+func (r *CheckoutProviderResult) MergeOnto(base *CheckoutProviderResult) *CheckoutProviderResult {
+	if r == nil {
+		return base
+	}
+	if base == nil {
+		return r
+	}
+
+	merged := *base
+	if r.NextAction != nil {
+		merged.NextAction = r.NextAction
+	}
+	if r.ProviderSessionID != "" {
+		merged.ProviderSessionID = r.ProviderSessionID
+	}
+	if r.ProviderPaymentIntentID != "" {
+		merged.ProviderPaymentIntentID = r.ProviderPaymentIntentID
+	}
+	if r.ExpiresAt != nil {
+		merged.ExpiresAt = r.ExpiresAt
+	}
+	if len(r.ProviderMetadata) > 0 {
+		if merged.ProviderMetadata == nil {
+			merged.ProviderMetadata = make(map[string]string, len(r.ProviderMetadata))
+		} else {
+			cloned := make(map[string]string, len(merged.ProviderMetadata)+len(r.ProviderMetadata))
+			for k, v := range merged.ProviderMetadata {
+				cloned[k] = v
+			}
+			merged.ProviderMetadata = cloned
+		}
+		for k, v := range r.ProviderMetadata {
+			merged.ProviderMetadata[k] = v
+		}
+	}
+	return &merged
+}
+
 func (r *CheckoutProviderResult) PaymentAction() *PaymentAction {
 	if r == nil {
 		return nil
--- a/internal/types/checkout_configuration.go
+++ b/internal/types/checkout_configuration.go
@@ -279,46 +279,102 @@ type CheckoutProviderResult struct {
 	ProviderMetadata map[string]string `json:"provider_metadata,omitempty"`
 }
 
-// MergeOnto returns a copy of base with the non-empty fields of r overlaid. Callers
-// that learn only part of the result — a webhook knowing the payment id but not the
-// redirect action — pass a fragment; writing it straight to the column would drop the
-// NextAction and ProviderSessionID recorded at link creation. Either side may be nil.
-func (r *CheckoutProviderResult) MergeOnto(base *CheckoutProviderResult) *CheckoutProviderResult {
-	if r == nil {
-		return base
-	}
-	if base == nil {
-		return r
-	}
+// CheckoutProviderResultBuilder assembles a CheckoutProviderResult, optionally on top
+// of one already stored.
+//
+// Most callers only learn part of the result — a webhook knows the gateway payment id
+// but not the redirect action recorded at link creation — so the value that gets
+// persisted has to combine what a caller brings with what is already there. Writing a
+// caller's fragment straight to the column would drop the NextAction and
+// ProviderSessionID, leaving no trace back to the provider object.
+type CheckoutProviderResultBuilder struct {
+	result CheckoutProviderResult
+}
 
-	merged := *base
+// NewCheckoutProviderResult starts an empty builder.
+func NewCheckoutProviderResult() *CheckoutProviderResultBuilder {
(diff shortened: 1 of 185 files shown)


internal/types/checkout_configuration.go
```diff
@@ -279,6 +279,48 @@ type CheckoutProviderResult struct {
 	ProviderMetadata map[string]string `json:"provider_metadata,omitempty"`
 }
 
+// MergeOnto returns a copy of base with the non-empty fields of r overlaid. Callers
+// that learn only part of the result — a webhook knowing the payment id but not the
+// redirect action — pass a fragment; writing it straight to the column would drop the
+// NextAction and ProviderSessionID recorded at link creation. Either side may be nil.
+func (r *CheckoutProviderResult) MergeOnto(base *CheckoutProviderResult) *CheckoutProviderResult {
```
Harshit-1104: Let's create a builder of CheckoutProviderResult

Harshit-1104 (changes requested):

Harshit-1104 (dismissed):

Harshit-1104 (approved):

PR #2762, opened 2026-09-03
feat(invoice): edit finalized invoices via void-and-recreate draft flow

dskhairnar:
## Summary

Backend for the finalized-invoice edit flow (frontend: flexprice/flexprice-front#1358). Editing a FINALIZED invoice now voids it and recreates it as a DRAFT copy, with the edit landing on the copy — **same endpoints, same request/response shapes**, so clients change nothing.

### What triggers the flow
- `POST /invoices/:id/modify/execute` (line item add/update/remove) on a FINALIZED invoice
- `PUT /invoices/:id` with `apply_discount: true` on a FINALIZED invoice

Plain field updates (due date, PDF URL, metadata) keep editing finalized invoices in place — only financial edits trigger the recreate.

### What the draft copy carries
All invoice data: **description**, **billing period + period start/end (the interval)**, due date, issue date, metadata, PDF URL, subtotal/discount/tax — and every published line item with its full fields (display name, amount, quantity, price/meter linkage, line-level period start/end). Payment state does not carry over: voiding refunds any captured payment / applied credits, so the draft starts clean (PENDING, amount_paid 0).

### Robustness
- Copied line items link to originals via `parent_line_item_id`; ops that reference the ORIGINAL invoice's line item ids are translated onto the copy.
- The voided original links to the copy via `recalculated_invoice_id`; a call (or retry) still targeting the voided original is redirected to the replacement draft — a finalized invoice is only ever recreated once (deterministic idempotency key on the copy).
- Responses return the draft so clients can chain subsequent calls to its id (the FE does).
- If the copy fails after voiding, the invoice is left voided with no replacement (logged for manual retry) — consistent with the existing RecalculateInvoice contract.

### Tests
- Modify-on-finalized: response is a new draft, original voided + linked; description/billing period/period start-end/due date/metadata/line-item periods all verified on the copy; the update landed on the copied item.
- Remove using the original invoice's line item ids (lineage translation).
- Second call targeting the voided original redirects to the same draft (no double recreate).
- `apply_discount` on finalized voids and recreates via PUT; an unvoidable invoice (e.g. REFUNDED) still rejects.
- Full `internal/ee/service` package green; `go build ./internal/... ./cmd/...` and `go vet` clean; swagger regenerated with the updated endpoint description.

## Summary by CodeRabbit

* **New Features**
  * Added APIs to list, retrieve, and retry refunds.
  * Added webhook events for refund creation, success, and failure.
  * Added refund targeting options to credit note workflows.
  * Added alert threshold type configuration.
  * Finalized invoices can now be modified by creating a replacement draft invoice.

* **Updates**
  * Replaced `customer_present` with `customer_not_present` in payment configuration.
  * Expanded invoice and payment webhook documentation.
  * Edits targeting voided invoices are rejected with the replacement invoice identified.


internal/ee/service/invoice_line_item_edit.go
```diff
@@ -261,21 +261,50 @@ func (s *invoiceService) ModifyInvoice(ctx context.Context, invoiceID string, re
...
+		lineItemIDs := make([]string, 0, len(params.LineItemIDs))
+		for _, id := range params.LineItemIDs {
+			resolved, resolveErr := s.resolveLineItemIDForEdit(ctx, targetID, id)
+			if resolveErr != nil {
+				return nil, resolveErr
+			}
+			lineItemIDs = append(lineItemIDs, resolved)
+		}
+		resp, err = s.RemoveBulkLineItem(ctx, targetID, dto.RemoveBulkLineItemRequest{
+			LineItemIDs: lineItemIDs,
```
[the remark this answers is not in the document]
dskhairnar: Fixed in d728057aa — referenced line item ids are now validated against the invoice's published items BEFORE voiding, so a bad id fails the request while the finalized invoice is intact. (The validator lists items directly rather than trusting invoice hydration.)

internal/ee/service/invoice_line_item_edit.go
```diff
@@ -288,3 +317,127 @@ func (s *invoiceService) executeLineItemModification(ctx context.Context, invoic
...
+		InvoiceStatus:          types.InvoiceStatusDraft,
+		PaymentStatus:          types.PaymentStatusPending,
+		Currency:               voided.Currency,
+		// Payment and credit state does not carry over: voiding refunded any captured
+		// payment and applied credits back to the customer, so the draft starts clean.
+		AmountPaid:                 decimal.Zero,
+		TotalPrepaidCreditsApplied: decimal.Zero,
+		Subtotal:                   voided.Subtotal,
+		TotalDiscount:              voided.TotalDiscount,
+		TotalTax:                   voided.TotalTax,
```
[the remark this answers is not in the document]
dskhairnar: Fixed in d728057aa — the copy no longer carries TotalDiscount/TotalTax, since their backing records (coupon applications, tax-applied rows) reference the voided invoice and are not copied. Both are re-derived on the draft: apply_discount recalculates from current coupon associations, and taxes are recalculated at finalize.

internal/ee/service/invoice_line_item_edit.go
```diff
@@ -288,3 +317,127 @@ func (s *invoiceService) executeLineItemModification(ctx context.Context, invoic
...
+		EnvironmentID:              voided.EnvironmentID,
+		// Deterministic key: a retried recreate resolves to the same replacement instead
+		// of racing a duplicate (idempotency lookups exclude VOIDED invoices).
+		IdempotencyKey:   lo.ToPtr("void_recreate-" + voided.ID),
+		IsManuallyEdited: true,
+		Version:          1,
+		BaseModel:        baseModel,
+		LineItems:        lineItems,
+	}
+	s.recalculateTotalsFromLineItems(draft, lineItems)
```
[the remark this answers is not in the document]
dskhairnar: Fixed in d728057aa — when the finalized invoice has no published line items, the copy keeps the original Subtotal (Total/AmountDue/AmountRemaining derived from it) instead of recomputing to zero. The apply_discount test now asserts the draft totals are preserved (100).

internal/ee/service/invoice.go
```diff
@@ -3825,6 +3825,27 @@ func (s *invoiceService) UpdateInvoice(ctx context.Context, id string, req dto.U
...
+		switch {
+		case inv.InvoiceStatus == types.InvoiceStatusVoided && inv.RecalculatedInvoiceID != nil:
+			id = *inv.RecalculatedInvoiceID
+		case inv.InvoiceStatus == types.InvoiceStatusFinalized:
+			draft, err := s.voidAndRecreateDraftForEdit(ctx, inv)
+			if err != nil {
+				return nil, err
+			}
+			id = draft.ID
+		}
```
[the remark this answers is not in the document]
dskhairnar: Fixed in d728057aa — both entry points now resolve the target through a shared followReplacementChain helper that walks recalculated_invoice_id links until the target is no longer voided (bounded at 10 hops). Covered by a new multi-hop test: an edit targeting the first original resolves across two hops to the current draft.

docs/swagger/docs.go
```diff
@@ -7218,6 +7226,244 @@ const docTemplate = `{
...
+                        "description": "Offset",
+                        "name": "offset",
+                        "in": "query"
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "OK",
+                        "schema": {
+                            "$ref": "#/definitions/ListRefundsResponse"
```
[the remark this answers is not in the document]
dskhairnar: Out of scope for this PR: the RefundFilter code is untouched here — this diff region is the regenerated swagger spec picking up annotations already on develop. Worth a separate issue/PR for bounding the id arrays.

docs/swagger/swagger.json
```diff
@@ -13350,8 +13665,8 @@
                 "collection_method": {
                     "$ref": "#/definitions/types.CollectionMethod"
                 },
-                "customer_present": {
-                    "description": "CustomerPresent declares the charge as customer-initiated (CIT) rather than\nmerchant-initiated (MIT). It changes which SCA/3DS exemptions the gateway may\nclaim and who carries chargeback liability, so it must describe reality: true\nonly when the customer is actually at the keyboard, as in a portal one-click\ntop-up. Unattended charges (auto top-up, dunning) leave it false.",
+                "customer_not_present": {
+                    "description": "CustomerNotPresent is the unattended/MIT opt-in. Zero value (omitted) means\nthe customer is present, so a missed auto-charge may fall back to a hosted\nauthorization link. Set true only from merchant-initiated paths (auto top-up).",
```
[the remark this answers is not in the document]
dskhairnar: Out of scope for this PR: the checkout session binding is untouched here — this diff region is the regenerated swagger spec reflecting changes already merged to develop. The customer_present migration question belongs to the PR that introduced CustomerNotPresent.

internal/ee/service/invoice_line_item_edit.go
```diff
@@ -288,3 +323,200 @@ func (s *invoiceService) executeLineItemModification(ctx context.Context, invoic
 
 	return &dto.InvoiceModifyResponse{Invoice: resp}, nil
 }
+
+// followReplacementChain resolves a voided invoice to its current replacement by
+// following recalculated_invoice_id links — a replacement draft can itself be
+// finalized, edited, and voided again, giving the chain more than one hop.
+func (s *invoiceService) followReplacementChain(ctx context.Context, inv *invoice.Invoice) (*invoice.Invoice, error) {
```
DiabolusGX: I think we can remove this.
If I'm trying to edit an invoice which is already re-calculated, we can still allow to edit it.

internal/ee/service/invoice_line_item_edit.go
```diff
@@ -288,3 +323,200 @@ func (s *invoiceService) executeLineItemModification(ctx context.Context, invoic
...
+// addressable. The original is linked to the copy via recalculated_invoice_id.
+func (s *invoiceService) voidAndRecreateDraftForEdit(ctx context.Context, inv *invoice.Invoice) (*invoice.Invoice, error) {
+	if inv.RecalculatedInvoiceID != nil {
+		return nil, ierr.NewError("invoice already has a replacement").
+			WithHintf("invoice %s was already voided and recreated as %s", inv.ID, *inv.RecalculatedInvoiceID).
+			WithReportableDetails(map[string]any{"recalculated_invoice_id": *inv.RecalculatedInvoiceID}).
+			Mark(ierr.ErrValidation)
+	}
+
+	if err := s.VoidInvoice(ctx, inv.ID, dto.InvoiceVoidRequest{}); err != nil {
```
DiabolusGX: can get updated void invoice in response

internal/ee/service/invoice_line_item_edit.go
```diff
@@ -261,21 +261,56 @@ func (s *invoiceService) ModifyInvoice(ctx context.Context, invoiceID string, re
...
+	}
+	inv, err = s.followReplacementChain(ctx, inv)
+	if err != nil {
+		return nil, err
+	}
+	targetID := inv.ID
+	if inv.InvoiceStatus == types.InvoiceStatusFinalized {
+		// Validate the referenced line items BEFORE voiding: a bad id must fail the
+		// request while the finalized invoice is still intact, not after the void.
+		if err := s.validateLineItemRefsAgainstInvoice(ctx, inv, params); err != nil {
```
DiabolusGX: don't need this either. we can simply start transaction here, acquire lock, map existing line items to new line items (invalid ones will automatically be filtered out)

internal/ee/service/invoice_line_item_edit.go
dskhairnar: Removed in 88b97e074 — the guard is gone. An invoice that already has a replacement follows the recalculated_invoice_id chain (followReplacementChain) and the edit lands on the current draft, so editing an already-recalculated invoice just works.

internal/ee/service/invoice_line_item_edit.go
dskhairnar: Done in 88b97e074 — VoidInvoice now returns the updated (voided) invoice, and both this flow and RecalculateInvoice use the returned value instead of re-reading. (interfaces.InvoiceService and all callers updated.)

internal/ee/service/invoice_line_item_edit.go
dskhairnar: Done in 88b97e074 — the pre-validation is gone. The finalized edit now runs in one transaction: GetForUpdate lock, void, recreate (building an old→new line item id map), translate the op through the map, apply. Stale remove ids are silently dropped; a targeted update of an unknown id errors and the transaction rolls the void and the copy back together. One note: VoidInvoice's post-tx side effects (void webhook, wallet cleanup) fire before the outer commit, so a rolled-back edit can still emit a void webhook — flagging in case that matters.

internal/ee/service/invoice.go
```diff
@@ -3825,9 +3821,46 @@ func (s *invoiceService) UpdateInvoice(ctx context.Context, id string, req dto.U
...
+			return nil, err
+		}
+		voidAndRecreate = inv.InvoiceStatus == types.InvoiceStatusFinalized
+	}
+
 	var updatedInv *invoice.Invoice
 	err := s.DB.WithTx(ctx, func(txCtx context.Context) error {
-		inv, err := s.InvoiceRepo.GetForUpdate(txCtx, id)
+		targetID := id
+		if voidAndRecreate {
```
DiabolusGX:
```suggestion
		if inv.InvoiceStatus == types.InvoiceStatusFinalized && req.ApplyDiscount {
					draft, _, err := s.voidAndRecreateDraftForEdit(txCtx, locked)
```

internal/ee/service/invoice_line_item_edit.go
```diff
@@ -261,30 +250,251 @@ func (s *invoiceService) ModifyInvoice(ctx context.Context, invoiceID string, re
...
-	case dto.InvoiceModifyLineItemActionRemove:
-		resp, err = s.RemoveBulkLineItem(ctx, invoiceID, dto.RemoveBulkLineItemRequest{
-			LineItemIDs: params.LineItemIDs,
+	case dto.InvoiceModifyLineItemActionUpdate, dto.InvoiceModifyLineItemActionRemove:
+		resolve, err := s.lineItemResolverForInvoice(ctx, invoiceID)
+		if err != nil {
+			return nil, err
+		}
+		if params.Action == dto.InvoiceModifyLineItemActionUpdate {
+			return s.UpdateLineItem(ctx, invoiceID, resolve(params.LineItemID), *params.Update)
```
DiabolusGX:
```suggestion
			return s.UpdateLineItem(ctx, invoiceID, params.LineItemID, *params.Update)
```

internal/ee/service/invoice_line_item_edit.go
```diff
@@ -261,30 +250,251 @@ func (s *invoiceService) ModifyInvoice(ctx context.Context, invoiceID string, re
...
 			WithHint("valid values: add, update, remove").
 			Mark(ierr.ErrValidation)
 	}
+}
+
+// lineItemResolverForInvoice lists the invoice's published line items once and returns a
+// pure resolver: an id already on the invoice passes through, an id whose copy points at
+// it via parent_line_item_id resolves to the copy, and anything else passes through
+// unchanged for the operation itself to reject.
+func (s *invoiceService) lineItemResolverForInvoice(ctx context.Context, invoiceID string) (func(string) string, error) {
```
DiabolusGX: remove this please

internal/ee/service/invoice_line_item_edit.go
```diff
@@ -261,30 +261,247 @@ func (s *invoiceService) ModifyInvoice(ctx context.Context, invoiceID string, re
...
+// voidAndRecreateDraftForEdit implements the finalized-invoice edit flow: it voids the
+// finalized invoice (refunding any captured payment or applied credits) and creates a DRAFT
+// copy carrying the invoice's data and every published line item. Copied line items point at
+// their originals via parent_line_item_id and the returned map translates old ids to new
+// ones. Discounts are re-derived on the copy from the subscription's current coupon
+// associations; taxes are recalculated when the draft is finalized. The original links to
+// the copy via recalculated_invoice_id. Callers run this inside a transaction so a failed
+// edit rolls the void back.
+func (s *invoiceService) voidAndRecreateDraftForEdit(ctx context.Context, inv *invoice.Invoice) (*invoice.Invoice, map[string]string, error) {
+	voided, err := s.VoidInvoice(ctx, inv.ID, dto.InvoiceVoidRequest{})
```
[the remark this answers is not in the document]
dskhairnar: Fixed in dd4595db6 — VoidInvoice now skips its side effects (voided webhook, wallet cleanup) when it runs inside an outer transaction (TxFromContext check); the edit flows run runVoidSideEffects only after their WithTx commits, so a rolled-back edit never announces a void.

internal/domain/invoice/draft_copy.go
```diff
@@ -0,0 +1,48 @@
...
+		WithID(id).
+		WithInvoiceStatus(types.InvoiceStatusDraft).
+		WithPaymentStatus(types.PaymentStatusPending).
+		WithAmountPaid(decimal.Zero).
+		WithTotalPrepaidCreditsApplied(decimal.Zero).
+		WithRefundedAmount(decimal.Zero).
+		WithAdjustmentAmount(decimal.Zero).
+		WithTotalDiscount(decimal.Zero).
+		WithTotalTax(decimal.Zero).
+		WithTaxExemptionReasonCode(nil).
```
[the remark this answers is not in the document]
dskhairnar: Fixed in dd4595db6 — CopyForDraftEdit no longer clears TaxExemptionReasonCode; the copy carries it so recreated one-off/credit invoices keep their exemption reason.

internal/ee/service/invoice.go
dskhairnar: Applied in dd4595db6 — UpdateInvoice now takes one GetForUpdate up front and branches exactly as suggested: FINALIZED + apply_discount → void-and-recreate (discount applied to the copy), DRAFT + apply_discount → in-place recalculation, voided → hard stop.

internal/ee/service/invoice_line_item_edit.go
dskhairnar: Applied in dd4595db6 — applyLineItemAction is pure dispatch now; stale ids resolve via parent lineage inside UpdateLineItem/RemoveBulkLineItem against the locked invoice's already-loaded line items (zero extra reads).

internal/ee/service/invoice_line_item_edit.go
dskhairnar: Removed in dd4595db6 — lineItemResolverForInvoice is gone, along with the id-map translation pass.

DiabolusGX (approved):

aurumflux20:
Read this one closely — void-and-recreate for finalized invoices is the kind of change where a quiet mistake becomes a double bill — and the merge commit's checks are clean: Run Tests and codecov/patch both green. That is more than most repos merging agent-assisted billing changes can show, so this is a question, not a report.

When the agent both changes the invoice logic and writes the tests for it in the same session, what checks that the tests were not shaped to fit the change? CI can only confirm that whatever tests exist pass; it cannot see whether "tests pass" was earned or arranged. That gap lives in the session, not in the diff.

We read agent sessions for exactly that — each claim against what was actually run — and sign the result so someone with no stake in it can verify it without trusting us: https://github.com/aurumflux20/coherence. If there is a session behind this PR you would want graded, say so and I will say what that involves. If not, the question stands on its own and no reply is owed.

(Tried tech@flexprice.io first; it bounced as a closed group, so here instead.)

PR #2774, opened 2026-09-05
feat(invoice): description and period interval on invoice line items

dskhairnar:
Adds `description` and `period_start`/`period_end` support to manually edited invoice line items (frontend: description text field + service-period range picker on the invoice edit page).

- **No schema change**: line item descriptions use the existing `metadata["description"]` convention — the same key the billing engine writes ("Base Plan (Fixed Charge)" etc.) and the invoice PDF renderer reads, so user-set descriptions render on the PDF automatically.
- `AddLineItemRequest` / `UpdateLineItemRequest` accept `description`, `period_start`, `period_end`. Reversed intervals are rejected — adds validate the request pair, updates validate the merged line item.
- Line item builder gains `WithPeriodStart` / `WithPeriodEnd` (periods are existing real columns), wired through `AddBulkLineItem` and `UpdateLineItem`.
- Round-trip test: add with description+interval, update shifting the interval, reversed-interval rejection on both paths. Full `ee/service` suite green; swagger regenerated.

## Summary by CodeRabbit

* **New Features**
  * Invoice line items now support descriptions and optional start and end dates when added or updated.
  * Partial updates preserve existing period values when they are not changed.

* **Bug Fixes**
  * Prevented invoice line items from being saved with an end date earlier than the start date.

* **Documentation**
  * Updated API documentation to reflect the new invoice line-item fields and enum details.

### DIFF
--- a/internal/ee/service/invoice_line_item_edit.go
+++ b/internal/ee/service/invoice_line_item_edit.go
@@ -32,18 +32,6 @@ func (s *invoiceService) recalculateTotalsFromLineItems(inv *invoice.Invoice, li
 	}
 }
 
-// setLineItemDescription writes the line item description into metadata["description"] —
-// the existing convention the billing engine writes and the invoice PDF reads.
-func setLineItemDescription(item *invoice.InvoiceLineItem, description *string) {
-	if description == nil {
-		return
-	}
-	if item.Metadata == nil {
-		item.Metadata = types.Metadata{}
-	}
-	item.Metadata["description"] = *description
-}
-
 func (s *invoiceService) UpdateLineItem(ctx context.Context, invoiceID, lineItemID string, req dto.UpdateLineItemRequest) (*dto.InvoiceResponse, error) {
 	if err := req.Validate(); err != nil {
 		return nil, err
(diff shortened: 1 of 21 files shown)


internal/ee/service/invoice_line_item_edit.go
```diff
@@ -32,6 +32,18 @@ func (s *invoiceService) recalculateTotalsFromLineItems(inv *invoice.Invoice, li
 	}
 }
 
+// setLineItemDescription writes the line item description into metadata["description"] —
+// the existing convention the billing engine writes and the invoice PDF reads.
+func setLineItemDescription(item *invoice.InvoiceLineItem, description *string) {
```
DiabolusGX: this method shuold take `builder` in input and call `setdescription` and `setmetadata` if we have to store description in metadata as well. not sure why we need this tho

dskhairnar: Done in 6eba89048 — it's a builder method now: WithDescription on invoiceLineItemBuilder, called in the build chain like every other setter. On why it exists at all: line items have no description column — the description must land in Metadata["description"], the key the billing engine writes ("Base Plan (Fixed Charge)" etc.) and the invoice PDF renderer reads. The builder method encodes that convention in one place instead of every caller knowing to poke the metadata map (and it merges into the seeded item's deep-copied metadata rather than replacing it, which a plain SetMetadata would clobber).

migrations/versioned/postgres/20260905000100_invoice_line_item_description.sql
```diff
@@ -0,0 +1,8 @@
+-- migrate:up
```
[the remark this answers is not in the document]
dskhairnar: Moot as of 05e814e25 — the migration (and the schema column) was removed entirely; descriptions use the existing metadata convention, so there is no migration to checksum.

DiabolusGX (approved):

PR #2786, opened 2026-09-07
ci(security): flip Semgrep + gosec to hard-fail on high-confidence rules

agrim123:
## What

Phase 2 of the SAST rollout in `sast.yml`. Both scanners now **block** the PR on findings instead of only reporting (soft-fail → hard-fail).

- **Semgrep** (`p/default` + `p/golang`): findings (exit 1) propagate and fail the job. Generated SDK output (`api/go`, `api/python`, `api/typescript`) is excluded — machine-generated from the OpenAPI spec, so suppressions there are overwritten on regeneration.
- **gosec**: scoped to `-severity high -confidence high` per the ticket ("high-confidence rules"); only high/high findings block.

## Baseline triage

Flipping surfaced findings the soft gate had been tolerating. Each was verified case-by-case (all confirmed non-security) and suppressed **per-site in config**, never by muting the tool:

| Finding | Sites | Verdict | Suppression |
|---|---|---|---|
| `math-random-used` | seed tooling, e2e probe, test data, Kafka dedup salts, workflow jitter | non-security (no tokens/keys/nonces) | `nosemgrep` on the **import line** (rule fires on the import); `#nosec G404` kept on call sites for gosec |
| `bypass-tls-verification` | `integration-testing-suite/go/http_client.go` | opt-in `insecure` test-client flag for preprod certs, never prod | `nosemgrep` w/ reason (`nolint:gosec` already present) |

Also removed a deprecated `rand.Seed` call (no-op since Go 1.20).

## Verification (local, on develop)

```
gosec -severity high -confidence high ./...          -> Issues: 0
semgrep p/default,p/golang (+ SDK excludes)          -> exit 0, 0 blocking findings
go build ./cmd/server + all touched packages         -> clean
```

## Follow-up (not code)

Add the **Semgrep** and **gosec** checks to `develop`/`main` **required status checks** in branch protection so the exit code blocks *merge*, not just the job. Same follow-up as #2785 (Trivy fs).

Closes FLE-1233

## Summary by CodeRabbit

* **Security**
  * Security scanning now blocks pull requests when applicable findings, scanner errors, or high-severity/high-confidence issues are detected.
  * Generated SDK content is excluded from scans, and suppressed findings are removed before results are uploaded.
  * Security-scanner annotations clarify approved uses of non-sensitive random data and optional insecure test configurations.

* **Maintenance**
  * Non-production event generation remains functional with simplified randomization behavior.

### DIFF
--- a/.github/workflows/sast.yml
+++ b/.github/workflows/sast.yml
@@ -99,5 +99,10 @@ jobs:
       # BLOCKING on high-confidence findings: scoped to -severity high
       # -confidence high so only high/high issues fail the PR. Findings (exit 1)
       # and processing errors (exit 2) both fail the job.
+      #
+      # api/custom and api/go are excluded: the custom SDK tree depends on
+      # generated types (*Flexprice) and does not compile standalone, so gosec's
+      # package load errors on it — a build-graph artifact, not a security issue.
+      # (Same generated-SDK exclusion as the Semgrep job.)
       - name: gosec scan
-        run: gosec -severity high -confidence high -fmt text ./...
+        run: gosec -severity high -confidence high -exclude-dir=api/custom -exclude-dir=api/go -fmt text ./...
--- a/.github/workflows/sast.yml
+++ b/.github/workflows/sast.yml
@@ -58,6 +58,18 @@ jobs:
             --sarif --output semgrep.sarif \
             .
 
+      # Drop results semgrep marked as suppressed (inline `nosemgrep`) before
+      # upload. semgrep keeps them in SARIF tagged `suppressions`, but GitHub
+      # code-scanning still posts each as a PR review comment, re-surfacing every
+      # triaged nosemgrep site. Stripping them keeps real findings visible in
+      # code-scanning without the dismissed-finding noise.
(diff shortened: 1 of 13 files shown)


agrim123:
Re: gosec excluding `api/custom/go` / `api/go`.

Acknowledged, but intentional and out of scope for this PR:

- `api/custom/go` + `api/go` are the **published Go SDK**, not the flexprice service. They depend on generated SDK types (`*Flexprice`) and do **not compile standalone** under this repo's gosec invocation — `gosec ./...` errors with `undefined: Flexprice` (package-load failure), which is what turned the newly-blocking gate red. That is a build-graph artifact, not a security finding (Issues: 0 either way).
- The SDK has its own generation + build pipeline (`make go-sdk` / `make merge-custom`) with the full type context; scanning it there is where an SDK gate belongs, not in the service SAST job.
- This PR (FLE-1233) only flips the existing service scanners soft→hard; adding a correctly-compiling SDK gosec step is separate work.

Filing the SDK-scan idea as a follow-up rather than blocking the gate flip on it. Resolving.

ojasagarwal-flex (approved):

PR #2810, opened 2026-09-08
feat(subscription): Line items grouping behavior

DiabolusGX:
## Summary by CodeRabbit

- **New Features**
  - Added subscription line-item grouping options: per charge period or per billing period.
  - Subscriptions can consolidate shorter-cadence charges into a single billing-period invoice line while preserving totals.
  - Added API support and validation for configuring this setting during subscription creation and updates.
  - Added subscription schedule listing, retrieval, and cancellation endpoints.
  - Enhanced checkout, invoice, custom-currency, payment, and synchronization API schemas.

- **Documentation**
  - Updated API reference documentation with new endpoints, fields, schemas, and supported grouping values.
  - Replaced legacy subscription-schedule route documentation with the current routes.


internal/types/subscription.go
```diff
@@ -200,6 +200,53 @@ func (c CollectionMethod) Validate() error {
...
+// cadence is shorter than the subscription's. Presentation only: charges are always
+// computed per charge period, so the total is the same under either value.
+type LineItemGrouping string
+
+const (
+	// LINE_ITEM_GROUPING_PER_CHARGE_PERIOD - one line item per charge period (monthly price on a quarterly sub bills as 3 rows)
+	LINE_ITEM_GROUPING_PER_CHARGE_PERIOD LineItemGrouping = "PER_CHARGE_PERIOD"
+
+	// LINE_ITEM_GROUPING_PER_BILLING_PERIOD - those rows collapsed into one spanning the billing period
+	LINE_ITEM_GROUPING_PER_BILLING_PERIOD LineItemGrouping = "PER_BILLING_PERIOD"
```
Harshit-1104: casing

internal/ee/service/billing.go
```diff
@@ -404,6 +405,133 @@ func applyOpeningInvoiceAdjustmentToLineItems(
...
+	// A fixed charge repeats the same quantity every window (5 seats each month),
+	// so summing it would report 15 seats for a 5-seat subscription.
+	if isUsageLineItem(target) {
+		target.Quantity = target.Quantity.Add(addition.Quantity)
+		target.AdjustedEntitlementQuantity = addDecimalPtr(target.AdjustedEntitlementQuantity, addition.AdjustedEntitlementQuantity)
+	}
+
+	if target.PriceUnitAmount != nil && addition.PriceUnitAmount != nil {
+		target.PriceUnitAmount = lo.ToPtr(target.PriceUnitAmount.Add(*addition.PriceUnitAmount))
+	}
```
Harshit-1104: unit amount will remain the same too right, why adding?
If it's the same charge, same unit amount would be there

internal/ee/service/billing.go
```diff
@@ -404,6 +405,133 @@ func applyOpeningInvoiceAdjustmentToLineItems(
...
+
+func laterTime(a, b *time.Time) *time.Time {
+	if a == nil {
+		return b
+	}
+	if b == nil || !b.After(*a) {
+		return a
+	}
+	return b
+}
```
Harshit-1104: i think such utils already exist, reuse them

internal/ee/service/billing.go
```diff
@@ -404,6 +405,133 @@ func applyOpeningInvoiceAdjustmentToLineItems(
...
+
+func addDecimalPtr(a, b *decimal.Decimal) *decimal.Decimal {
+	if a == nil {
+		return b
+	}
+	if b == nil {
+		return a
+	}
+	return lo.ToPtr(a.Add(*b))
```
Harshit-1104: no lo.Add or something?
feels like an unnecessary util

Harshit-1104 (changes requested):

internal/ee/service/billing.go
DiabolusGX: Those are across services, thought of using them but skipped. Will use them

internal/types/subscription.go
DiabolusGX: Done

internal/ee/service/billing.go
DiabolusGX: The name is misleading: it means "the amount, expressed in the price-unit currency," not "the per-unit amount."

internal/ee/service/billing.go
DiabolusGX: handled differently, `lo` doesn't have "Add" it's a mathematical operation on `Decimal` type, which is not primitive

Harshit-1104 (approved):

PR #2839, opened 2026-09-11
feat: add database read-only mode

agrim123:
## What

Adds a `postgres.readonly` config flag (env `FLEXPRICE_POSTGRES_READONLY`, default false) that puts the database into read-only mode.

When enabled:
- an ent mutation hook on the writer and reader clients rejects Create/Update/Delete with `ErrReadOnly`
- `RequirePermission` returns 503 on write-action routes (keyed on the RBAC action, not HTTP method — since read endpoints may use POST)

Reads, event ingestion (Kafka-only), and login are unaffected. The disabled path is a pure passthrough with negligible overhead. Default is false, so there is no behavior change until it is explicitly enabled.

## Testing

- `go build ./internal/... ./cmd/...` — clean
- `go test ./internal/config/ ./internal/postgres/ ./internal/rest/middleware/ ./internal/ee/service/ -count=1` — all pass

### DIFF
--- a/docs/specs/db-write-freeze-readonly-mode.md
+++ b/docs/specs/db-write-freeze-readonly-mode.md
@@ -1,0 +1,183 @@
+# Spec: DB write-freeze (read-only mode) for AlloyDB cutover
+
+## Goal
+
+A flag that freezes ALL Postgres writes while keeping reads, event ingestion,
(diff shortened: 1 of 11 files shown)


agrim123:
Closing — opened against the wrong (public) repo by mistake. Moving to the correct private location.

PR #2863, opened 2026-09-15
fix(onboarding): grant super_admin and wire caches in onboard-tenant script

ojasagarwal-flex:
## Problem

Running the `onboard-tenant` seed script (`OnboardNewTenant`) produced a **half-onboarded tenant** that could not use the app. Found while onboarding a real tenant on AWS India prod.

Two distinct bugs:

1. **No role granted.** The script created the first user via `user.NewUser` (`Roles: []`). Every RBAC-guarded request then failed with `insufficient permissions` — e.g. `PUT /v1/tenants/update`. The primary onboarding flow (`internal/ee/service/onboarding.go:441`) grants the first user `super_admin`; the script did not.

2. **Nil-cache panic.** `userRepository.Create` calls `DeleteCache` after insert, which dereferenced a **nil `redisCache`** (the script wired no cache into `RepositoryParams`) and panicked — after the Postgres user row committed but before the Sandbox environment loop ran. Result: tenant + user exist, but no environment and no role.

Downstream, the empty environment list made the frontend send a stale `x-environment-id` it could not self-heal, surfacing as `invalid or inaccessible environment`.

## Fix

- Wire `InMemoryCache` and `RedisCache` into the script's `RepositoryParams`.
- Grant `super_admin` to the created user, mirroring the primary onboarding flow.
- Nil-guard the user repository cache methods (`GetCache` / `SetCache` / `DeleteCache`) so seed tooling run without a reachable Redis degrades gracefully instead of panicking (`cache.GetRedisCache()` can return `nil`, so wiring alone is insufficient).

## Scope / follow-ups (not in this PR)

- **Idempotency:** re-running still creates a duplicate tenant and won't upgrade an existing user's roles.
- **RBAC cache staleness:** role changes don't invalidate the auth user-cache, so new permissions can lag up to the Redis TTL (30 min).
- **Frontend:** shouldn't send an `x-environment-id` absent from the fetched list; should heal when the list goes empty → non-empty.

## Testing

- `go build ./scripts/... ./internal/repository/...` and `go vet ./scripts/internal/...` — clean.
- Verified on AWS India prod: after the fix's equivalent state, `GET /v1/environments` and `PUT /v1/tenants/update` both return 200.

## Summary by CodeRabbit

- **New Features**
  - Added support for bearer-token and environment ID authentication in the TypeScript SDK.
  - Updated SDK generation to include the new authentication configuration.

- **Bug Fixes**
  - Tenant onboarding now assigns the first user the super administrator role.
  - User repository operations now continue safely when Redis caching is not configured.


internal/repository/ent/user.go
```diff
@@ -488,6 +494,9 @@ func (r *userRepository) GetCache(ctx context.Context, id string) *domainUser.Us
 }
 
 func (r *userRepository) DeleteCache(ctx context.Context, id string) {
+	if r.redisCache == nil {
```
DiabolusGX, line 497: can skip this for now, we can add redis client in RepositoryParams

